PR #104, opened 2022-03-24
Add a basic docc bundle for generating documentation

phausler:
This has a fancy pants work-around from @ethan-kusters and @franklinsch. It gives us a full top level intro with a listing of extensions first and then global functions next.

### DIFF
--- a/bin/update-gh-pages-documentation-site
+++ b/bin/update-gh-pages-documentation-site
@@ -1,0 +1,68 @@
+#!/bin/bash
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2022 Apple Inc. and the Swift project authors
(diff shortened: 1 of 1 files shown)


ethan-kusters:
@phausler we had to avoid the Swift-DocC Plugin dependency in the SymbolKit repo as well. Here's how we worked around the issue: https://github.com/apple/swift-docc-symbolkit/blob/main/bin/update-gh-pages-documentation-site.


Another alternative might be something like:

```swift
if ProcessInfo.processInfo.environment["BUILDING_FOR_DOCUMENTATION_GENERATION"] != nil {
    package.dependencies += [
        .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
    ]
}
```

phausler:
oh that is a good idea. I will tinker with that.

Issue #45, opened 2022-02-17
Performance tests for merge only measure one active side


phausler:
The test currently measures the performance of merge as such:

```
  func test_merge2() async {
    await measureSequenceThroughput(output: 1) {
      merge($0, (0..<10).async)
    }
  }
```

Ideally we should measure cases where we have two inputs continuously emitting values. This likely needs a second emitter being added into `measureSequenceThroughput` so that it can be used for zip and other such combination APIs.

kingreza:
I think we can close this one out.

PR #114, opened 2022-03-26
#45 - Two sided performance test for merge

kingreza:
This PR extends `measureSequenceThroughput` to take two outputs so functions like merge/ zip etc can be tested by both sides.

### DIFF
--- a/Tests/AsyncAlgorithmsTests/Performance/TestThroughput.swift
+++ b/Tests/AsyncAlgorithmsTests/Performance/TestThroughput.swift
@@ -15,13 +15,13 @@ import AsyncAlgorithms
 #if canImport(Darwin)
 final class TestThroughput: XCTestCase {
   func test_chain2() async {
-    await measureSequenceThroughput(output: 1) {
-      chain($0, [].async)
+    await measureSequenceThroughput(firstOutput: 1, secondOutput: 2) {
+      chain($0, $1)
     }
   }
   func test_chain3() async {
-    await measureSequenceThroughput(output: 1) {
-      chain($0, [].async, [].async)
+    await measureSequenceThroughput(firstOutput: 1, secondOutput: 2, thirdOutput: 3) {
+      chain($0, $1, $2)
     }
   }
   func test_compacted() async {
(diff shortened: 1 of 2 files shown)


Tests/AsyncAlgorithmsTests/Performance/TestThroughput.swift
```diff
@@ -24,6 +24,11 @@ final class TestThroughput: XCTestCase {
       chain($0, [].async, [].async)
     }
   }
+  func test_chain4() async {
```
phausler, line 27: I think the names of these tests were intended to be per the number of inputs, so this is a test of chain 2 (but two sides active). @kperryua that was your intent right?

kingreza: Ah I see, if that's the case and confirmed I can make the change.

kingreza: I updated the tests, I also extended `measureSequenceThroughput` to take three parameters to cover the three input sequence for `chain`, `merge` and `zip`

phausler (approved):

Tests/AsyncAlgorithmsTests/Performance/ThroughputMeasurement.swift
```diff
@@ -87,6 +87,57 @@ extension XCTestCase {
     }
   }
   
+  public func measureSequenceThroughput<S: AsyncSequence, Output>(firstOutput: @autoclosure () -> Output, secondOutput: @autoclosure () -> Output, _ sequenceBuilder: (InfiniteAsyncSequence<Output>, InfiniteAsyncSequence<Output>) -> S) async where S: Sendable {
+    let metric = _ThroughputMetric()
+    let sampleTime: Double = 0.1
+    
+    measure(metrics: [metric]) {
+      let firstInfSeq = InfiniteAsyncSequence(value: firstOutput(), duration: sampleTime)
+      let secondInfSeq = InfiniteAsyncSequence(value: secondOutput(), duration: sampleTime)
```
kperryua, line 96: I just realized that the `duration` property on `InfiniteAsyncSequence` is vestigial from prior evoluation of this code. But we can handle removing that in a separate PR.

kperryua (approved):

Issue #120, opened 2022-03-28
Sendable conformance on _AsyncBytesBuffer.Storage allows unsafe concurrent access


mhjacobson:
In AsyncBufferedByteIterator.swift, `_AsyncBytesBuffer.Storage` (a class) conforms to `Sendable`, allowing it to be passed across concurrency boundaries.  Since concurrent access is allowed, classes conforming to `Sendable` are responsible for implementing internal synchronization, but `Storage` doesn't have any.

This means that it's possible to construct data races on the buffer of an `_AsyncBytesBuffer.Storage` (through copies of an `AsyncBufferedByteIterator`).  (See [this gist](https://gist.github.com/mhjacobson/ad1a1dee71f3f0e50102d62b6af0d49f) for an example.)  Here's what happens:

1. Multiple instances of `_AsyncBytesBuffer` value can share a single `Storage`.  If the multiple instances are read from on separate threads, then their `readFunction`s may be called concurrently, with the same pointer argument.  Since the job of the `readFunction` is to fill in the pointee bytes, there is a write/write data race on those bytes

2. Similarly, one copy may be reading from the `Storage` while another is calling its `readFunction`, producing a read/write data race on the bytes

mhjacobson:
I'll note that, disregarding the data races, even non-concurrent use of multiple copies of an `AsyncBufferedBytesIterator` has maybe unexpected behavior.  If one copy reloads before other copies have reached the end of the buffer, the other copies will skip bytes.

Therefore, it's not clear to me what guarantees `AsyncBufferedBytesIterator` intends to make about use of copies in general.

However, I might suggest that if using multiple copies of an `AsyncBufferedBytesIterator` is discouraged, then perhaps `AsyncBufferedBytesIterator` could itself be made a reference type (at least, until the language supports move-only types).

One solution might be to make it an actor, which would fix the data races, too...

phausler:
actor in this case is way too expensive and will introduce another race to an out of order access.

PR #121, opened 2022-03-28
#120 Ensure concurrent access to `AsyncBufferedByteIterator` does not crash from dangling pointers

phausler:
This resolves issue #120 by ensuring the shared buffers are reloaded independently. This does not promise any data consistency across tasks, but does ensure that the iteration can be sent as a move from one task to another.

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncBufferedByteIterator.swift
+++ b/Sources/AsyncAlgorithms/AsyncBufferedByteIterator.swift
@@ -110,6 +110,24 @@ internal struct _AsyncBytesBuffer: @unchecked Sendable {
     }
     try Task.checkCancellation()
     do {
+      // If two tasks have access to this iterator then the references on
+      // the storage will be non uniquely owned. This means that any reload
+      // must happen into it's own fresh buffer. The consumption of those
+      // bytes between two tasks are inherently defined as potential
+      // duplication by the nature of sending that buffer across the two
+      // tasks - this means that the brief period in which they may be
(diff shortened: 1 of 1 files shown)


Sources/AsyncAlgorithms/AsyncBufferedByteIterator.swift
```diff
@@ -110,6 +110,24 @@ internal struct _AsyncBytesBuffer: @unchecked Sendable {
...
+      // edge case of making the iterator and sending that across to two
+      // places to iterate is asking for something bizzare and the answer
+      // should not be crash, but it definitely cannot be consistent.
+      //
+      // The unique ref check is here to prevent the potentials of a crashing
+      // secnario.
+      if !isKnownUniquelyReferenced(&storage) {
+        // The count is not mutated across invocations so the access is safe.
+        let capacity = storage.buffer.count
+        storage = Storage(capacity: capacity)
```
parkera, line 129: So - if it doesn't crash, but it is an inconsistent state, perhaps we should *always* crash instead, by asserting? What do we expect the behavior to actually be if two tasks are using the same iterator?

phausler: In raw technical terms this type is a move-only Sendable type. We don't have move only semantics yet.

parkera: Ok, thanks for the additional info.

kperryua (approved):

PR #125, opened 2022-03-29
Remove obsolete property from measureSequenceThroughput

kingreza:
`duration` and `start` property in `measureSequenceThroughput` is [vestigial from early development](https://github.com/apple/swift-async-algorithms/pull/114#discussion_r836947233). This PR cleans that up.


phausler (approved):
Thanks for the cleanup! looks great.

PR #111, opened 2022-03-26
Adjacent Pairs

tevelee:
Similarly to swift-algorithms package [introduced](https://github.com/apple/swift-algorithms/pull/119) `adjacentPairs()`, it's equally relevant in the async/await world. 

I took inspiration from their implementation and updated the relevant documentation.

PR #126, opened 2022-03-29
Follow-up on adjacent pairs

tevelee:
Noticed an issue in https://github.com/apple/swift-async-algorithms/pull/111, the file name was missing the `Async` prefix. The class name and the file reference in the [documentation](https://github.com/apple/swift-async-algorithms/blob/main/Guides/AdjacentPairs.md) seems to be correct.


kperryua (approved):
Good catch.

Issue #131, opened 2022-03-30
AsyncChannel and AsyncThrowingChannel can hang tasks when that task is cancelled


phausler:
```swift
let channel = AsyncChannel<Int>()
let task = Task {
  print("awaiting")
  await channel.send(1)
  print("done sending")
}
try! await Task.sleep(nanoseconds: 1_000_000_000)
task.cancel()
print("cancelled task")
```

https://forums.swift.org/t/leaking-tasks-if-a-channel-receiver-is-no-longer-listening/56330

PR #132, opened 2022-03-30
#131 Cancel sends as if they were terminal events on task cancellation

phausler:
This makes AsyncChannel and AsyncThrowing channel no longer cause tasks to hang when they are cancelled.


phausler:
cc @jonathanpenn

jonathanpenn:
🙌

kperryua (approved):

PR #133, opened 2022-03-30
#113 Define the first element of throttled sequences by not exceeding  the rate (and fix a expectation misalignment of events)

### LINKED ISSUE #113: throttle should define the behavior of the first element
I find there are significant flaws in current `throttle` implementation.

1. First value is not emitted
2. Value is not emitted at throttled-interval-end when `latest = true`

### 1. First value is not emitted

From test code:

```swift
  func test_rate_3() {
    validate {
      "abcdefghijk|"
      $0.inputs[0].throttle(for: .steps(3), clock: $0.clock)
      "--c--f--i--|"
    }
  }
```

AFAIK, this `throttle` behavior is not compatible with any existing Rx frameworks, e.g. Combine, RxSwift, ReactiveSwift.

The correct result should (at least) be:

```swift
  func test_rate_3() {
    validate {
      "abcdefghijk|"
      $0.inputs[0].throttle(for: .steps(3), clock: $0.clock)
      "a--d--g--j-|"
    }
  }
```

where first `a` is emitted correctly, then start throttling for intervals.

### 2. Value is not emitted at throttled-interval-end when `latest = true`

From existing framework's behavior, `latest = true` normally means the latest cached value will be emitted at "throttled-interval-end".

For example:

```swift
    func test() async throws {
        let now = ContinuousClock.now

        let countUpStream = AsyncTimerSequence(interval: .seconds(1), clock: .continuous)
            .map { now.duration(to: $0).components.seconds } // 1, 2, 3, ...

        let throttledStream = countUpStream.throttle(for: .seconds(2.1), latest: true)

        let task = Task {
            for try await value in countUpStream {
                print("===> count", value)
            }
        }

        for try await value in throttledStream {
            print("===> throttled", value)
        }

        _ = await task.result

        print("===> done")
    }
```

This test runs timer starting from `1 sec`, and throttling interval is `2.1 sec`.

When `latest = true`, the ideal result should print:

```
// Ideal result
===> count 1
===> count 2
===> throttled 2 // at 2.1 sec
===> count 3
===> count 4
===> throttled 4 // at 4.2 sec
===> count 5
===> count 6
===> throttled 6 // at 6.3 sec
===> count 7
===> count 8
===> throttled 8 // at 8.4 sec
```

But current impl results:

```
// Current result
===> count 1
===> count 2
===> count 3
===> throttled 3 // at 3 sec
===> count 4
===> count 5
===> count 6
===> throttled 6 // at 6 sec
===> count 7
===> count 8
===> count 9
===> throttled 9 // at 9 sec
```

phausler:
This resolves #113 by defining the first element as not exceeding the reduction of the duration. 

Fixing that caused a related issue with the validation reconstruction to show a mis-alignment of the events.

### DIFF
--- a/Guides/Throttle.md
+++ b/Guides/Throttle.md
@@ -77,6 +77,8 @@ extension AsyncThrottleSequence.Iterator: Sendable
 
 The `AsyncThrottleSequence` and its `Iterator` are conditionally `Sendable` if the base types comprising it are `Sendable`.
 
+The time in which events are measured are from the previous emission if present. If a duration has elapsed between the last emission and the point in time the throttle is measured then that duration is counted as elapsed. The first element is considered not throttled because no interval can be constructed from the start to the first element.
+
 ## Alternatives Considered
 
 It was considered to only provide the "latest" style APIs, however the reduction version grants more flexibility and can act as a funnel to the implementations of `latest`.
--- a/Sources/AsyncAlgorithms/AsyncThrottleSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncThrottleSequence.swift
@@ -64,6 +64,7 @@ extension AsyncThrottleSequence: AsyncSequence {
   /// The iterator for an `AsyncThrottleSequence` instance.
   public struct Iterator: AsyncIteratorProtocol {
     var base: Base.AsyncIterator
+    var last: C.Instant?
     let interval: C.Instant.Duration
     let clock: C
     let reducing: @Sendable (Reduced?, Base.Element) async -> Reduced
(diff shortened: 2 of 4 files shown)


parkera:
Let's update the [guide](https://github.com/apple/swift-async-algorithms/blob/main/Guides/Throttle.md) with this info too.

Does this address the second part of that bug report (which we could split into a separate bug, if needed)?

phausler:
It handles the second part because it is measuring from the last emission not the start of iteration.

parkera (approved):

PR #135, opened 2022-04-02
Moves `XCTestCase.validate` into its own module

### LINKED ISSUE #62: `XCTestCase.validate` should be split off into its own module for use outside of `AsyncAlgorithms` tests
`XCTestCase.validate` should be split off into its own module for use outside of `AsyncAlgorithms` tests. The module should be called `AsyncAlgorithms_XCTest` so that perhaps we might be able to take advantage of [cross-import overlays](https://gist.github.com/beccadax/8f35216340e829efa5e9c1fc090fda09) when/if that becomes a ratified thing for swift packages.

kristofferjohansson:
Resolves #62 by moving the validate extension on XCTestCase to its own module.


Tests/AsyncAlgorithmsTests/TestValidationTests.swift
```diff
@@ -12,6 +12,7 @@
 import XCTest
 import AsyncAlgorithms
 import AsyncSequenceValidation
+@testable import AsyncAlgorithms_XCTest
```
phausler, line 15: so this prevents us from doing release-mode testing of perf. I am of the belief that this DEFINITELY should be tested, but we kinda need to settle on a way to do this AND accommodate for testing performance too

Tests/AsyncAlgorithmsTests/TestThrottle.swift
```diff
@@ -11,6 +11,7 @@
 
 import XCTest
 import AsyncAlgorithms
+import AsyncAlgorithms_XCTest
```
phausler, line 14: I presume this is needed for the clock tests eh?

kristofferjohansson:
> Overall this looks great, I presume that it works fine for importing to other project testing on both Darwin and Linux?

Thanks! I have only tested importing it to a test target in another package on Darwin so far. I can try to test it on Linux as well.

Tests/AsyncAlgorithmsTests/TestValidationTests.swift
kristofferjohansson: I actually didn't realize that this prevents that. I thought about moving the tests to their own test target for the new module, but they depend on other types in the `AsyncAlgorithmsTests` module and I tried to make as few changes as possible. I guess making more of the methods in the extension doesn't feel like a good solution either?

Tests/AsyncAlgorithmsTests/TestThrottle.swift
kristofferjohansson: This must have been a miss on my part, I have pushed a commit that removes this.

phausler:
was able to test this out on linux, seems like it works just fine!

phausler (approved):

PR #137, opened 2022-04-03
Add docc support and publish documentation to GitHub pages

Kyle-Ye:
- Move Guides into swift-docc based format
- Publish documentation to Github pages

> https://github.com/apple/swift-docc/pull/109


Package.swift
```diff
@@ -15,7 +15,9 @@ let package = Package(
     .library(name: "AsyncSequenceValidation", targets: ["AsyncSequenceValidation"]),
     .library(name: "_CAsyncSequenceValidationSupport", type: .static, targets: ["AsyncSequenceValidation"])
   ],
-  dependencies: [],
+  dependencies: [
+    .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
```
phausler, line 19: one idea that might be good here is that we conditionally do this by environment parameter; reducing the porting cost to embedded systems.

e.g. if we add the following to the top of the Package.swift
```
#if canImport(Darwin)
import Darwin
let buildingDocs = getenv("BUILDING_FOR_DOCUMENTATION_GENERATION") != nil
#elseif canImport(Glibc)
import Glibc
let buildingDocs = getenv("BUILDING_FOR_DOCUMENTATION_GENERATION") != nil
#else
let buildingDocs = false
#endif

// Only require the docc plugin when building documentation
let dependencies: [Package.Dependency] = buildingDocs ? [
  .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
] : []
```

and then change this to `dependencies: dependencies,`

and then `export BUILDING_FOR_DOCUMENTATION_GENERATION=1` from the doc build script

bin/update-gh-pages-documentation-site
```diff
@@ -0,0 +1,86 @@
...
+ASYNC_SEQUENCE_VALIDATION_OUTPUT_DIR="$ASYNC_ALGORITHMS_BUILD_DIR"/async-sequence-validation-docs
+
+mkdir -p "$ASYNC_SEQUENCE_VALIDATION_OUTPUT_DIR"
+
+# Set current directory to the repository root
+cd "$SWIFT_ASYNC_ALGORITHMS_ROOT"
+
+# Use git worktree to checkout the gh-pages branch of this repository in a gh-pages sub-directory
+git fetch
+git worktree add --checkout gh-pages origin/gh-pages
```
phausler, line 38: to lands this we will need to make a blank branch `gh-pages` as well as do the remaining setup for that

bin/update-gh-pages-documentation-site
```diff
@@ -0,0 +1,86 @@
...
+    --transform-for-static-hosting \
+    --hosting-base-path swift-async-algorithms \
+    --output-path "$SWIFT_ASYNC_ALGORITHMS_ROOT/gh-pages/docs"
+
+# Generate documentation for the 'AsyncSequenceValidation' target and output it
+# to a temporary output directory in the .build directory.
+swift package \
+    --allow-writing-to-directory "$ASYNC_ALGORITHMS_BUILD_DIR" \
+    generate-documentation \
+    --target AsyncSequenceValidation \
```
phausler, line 59: since this target is solely a testing artifact perhaps we should just skip it for now.

phausler (commented):
@parkera and I discussed earlier about how this does mean that we are publishing docs before stuff is completely settled per reviews - as long as we generate stuff at given sub-major milestones before 1.0 I think that should be fine as per documentation.

bin/update-gh-pages-documentation-site
Kyle-Ye: Got it, I’ll update it.

bin/update-gh-pages-documentation-site
Kyle-Ye: Yes, we need a `gh-pages` branch and add the environment GitHub action.

Package.swift
Kyle-Ye:
```swift
#if canImport(Darwin)
import Darwin
let buildingDocs = getenv("BUILDING_FOR_DOCUMENTATION_GENERATION") != nil
#elseif canImport(Glibc)
import Glibc
let buildingDocs = getenv("BUILDING_FOR_DOCUMENTATION_GENERATION") != nil
```
Can this be simplified to be the following?
```swift
import Foundation
let buildingDocs = ProcessInfo.processInfo.environment["BUILDING_FOR_DOCUMENTATION_GENERATION"] == nil
```
One question is do we need to `#if canImport(Foundation)`. I’m not quiet familiar with the current’s state of swift for embedded systems support.

phausler: Technically I also wanted this project to have a potential at offering bits to Foundation too; since the `AsyncBufferedByteIterator` is the backing implementation for `FileHandle.AsyncBytes`. Making it use just getenv poses a singular requirement on a common ancestor to all potential projects.

Kyle-Ye:
Fixed the CR suggestion and rebased main to solve the conflicts.
Could you please review this again? @phausler

phausler (approved):
So ive been coordinating with some other folks to get the remainder of the back-end tasks finished for this. There are some additional tasks we will need before it will all be ready to publish;

1) we need a redirect similar to how swift-argument-parser did it (however perhaps it might be good to stash that redirect.html in say a `lib/` directory.
2) some of the sub-titles are kinda wonky for the guides being imported as articles some minor restructuring will be needed for those.
3) I think there are some general enhancements we can do for the documentation to show methods as extensions to `AsyncSequence` (I have some ideas here we can explore)

none of those are to say blockers for getting this first step done; we can split it up into a couple of issues so we can get it done together.

phausler:
here is what it looks like currently: https://phausler.github.io/swift-async-algorithms/documentation/asyncalgorithms/

PR #141, opened 2022-04-09
Fix redundant where clauses in `AsyncBufferSequence` extension

TiagoMaiaL:
# Goal

To fix two warnings regarding redundant conformance constraints in `AsyncBufferSequence`:

Screen Shot 2022-04-08 at 21 25 57

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncBufferSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncBufferSequence.swift
@@ -234,8 +234,8 @@ public struct AsyncBufferSequence<Base: AsyncSequence, Buffer: AsyncBuffer> wher
   }
 }
 
-extension AsyncBufferSequence: Sendable where Base: Sendable, Base.AsyncIterator: Sendable, Base.Element: Sendable { }
-extension AsyncBufferSequence.Iterator: Sendable where Base: Sendable, Base.AsyncIterator: Sendable, Base.Element: Sendable { }
+extension AsyncBufferSequence: Sendable where Base: Sendable { }
+extension AsyncBufferSequence.Iterator: Sendable where Base: Sendable { }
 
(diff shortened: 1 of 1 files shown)


Sources/AsyncAlgorithms/AsyncBufferSequence.swift
```diff
@@ -234,8 +234,8 @@ public struct AsyncBufferSequence<Base: AsyncSequence, Buffer: AsyncBuffer> wher
   }
 }
 
-extension AsyncBufferSequence: Sendable where Base: Sendable, Base.AsyncIterator: Sendable, Base.Element: Sendable { }
-extension AsyncBufferSequence.Iterator: Sendable where Base: Sendable, Base.AsyncIterator: Sendable, Base.Element: Sendable { }
+extension AsyncBufferSequence: Sendable where Base: Sendable { }
```
phausler, line 237: @kperryua do you think we should just make `AsyncBufferSequence` require the `Base` to be `Sendable` and avoid this type of thing?

TiagoMaiaL: Just to be clear, the warnings are specific to the `Base.Element: Sendable` constraint. I took the liberty to also remove the `Base.AsyncIterator` constraint as well.

The declaration of `AsyncBufferSequence` already defines them:
```swift
public struct AsyncBufferSequence<Base: AsyncSequence, Buffer: AsyncBuffer> where Base.Element == Buffer.Input, Base.AsyncIterator: Sendable {  }
```
- It contains an explicit `AsyncIterator` constraint.
- `Base.Element` must be `Buffer.Input (aka AsyncBuffer.Input)`, which is already `Sendable` (line 130).

kperryua (approved):

PR #145, opened 2022-04-11
Add a redirect for the base page rendering

phausler:
This redirects rendered docs into the documentation category.

### DIFF
--- a/lib/redirect.html
+++ b/lib/redirect.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+   <head>
+      <title>ArgumentParser Documentation</title>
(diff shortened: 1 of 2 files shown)


lib/redirect.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+   <head>
+      <title>ArgumentParser Documentation</title>
```
TiagoMaiaL, line 4: Shouldn't this be AsyncAlgorithms Documentation?

phausler: oof oops, teaches me to copy paste :P

FranzBusch:
@phausler can we close this in favour of: https://github.com/apple/swift-async-algorithms/pull/297

PR #147, opened 2022-04-15
[WithLatestFrom] implement operator for arity 1 and 2

twittemb:
Hi.

This PR aims to propose a new operator: `withLatest(from:)`.
It has been discussed in the forum [here](https://forums.swift.org/t/pitch-withlatestfrom/56487/28).

I am new to the `evolution proposal process`. This PR contains:
- a proposal based on the sample + the guides.
- the implemented operator with unit tests and documentation.

Don't hesitate to guide me through the process 😀.

Thanks.

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncWithLatestFrom2Sequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncWithLatestFrom2Sequence.swift
@@ -1,0 +1,231 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+public extension AsyncSequence {
+  /// Combines `self` with two other ``AsyncSequence`` into a single ``AsyncSequence`` where each
+  /// element from `self` is aggregated to the latest known elements from the `other` sequences (if any) as a tuple.
+  ///
+  /// Remark: as the `other` sequences are being iterated over in the context of their own ``Task``, there is no guarantee
+  /// that their latest know elements are the ones that have just been produced when the base sequence produces its next element.
+  ///
+  /// ```
+  /// let base = AsyncChannel<Int>()
+  /// let other1 = AsyncChannel<String>()
+  /// let other2 = AsyncChannel<String>()
+  /// let sequence = base.withLatest(from: other1, other2)
+  ///
+  /// Task {
+  ///   for element in await sequence {
+  ///    print(element)
+  ///   }
+  /// }
+  ///
+  /// await other1.send("a")
+  /// await other1.send("b")
+  ///
+  /// await other2.send("c")
+  /// await other2.send("d")
+  ///
+  /// ... later in the application flow
+  ///
+  /// await base.send(1)
+  ///
+  /// // will print: (1, "b", "d")
+  /// ```
+  ///
+  /// - Parameters:
+  ///   - other1: the first other ``AsyncSequence``
+  ///   - other2: the second other ``AsyncSequence``
+  /// - Returns: an ``AsyncWithLatestFrom2Sequence`` where elements are a tuple of an element from `self` and the
+  /// latest known elements (if any) from the `other` sequences.
+  func withLatest<Other1: AsyncSequence, Other2: AsyncSequence>(
+    from other1: Other1,
+    _ other2: Other2
+  ) -> AsyncWithLatestFrom2Sequence<Self, Other1, Other2> {
+    AsyncWithLatestFrom2Sequence(self, other1, other2)
+  }
+}
+
+/// ``AsyncWithLatestFrom2Sequence`` is an ``AsyncSequence`` where elements are a tuple of an element from `base` and the
+/// latest known element (if any) from the `other` sequences.
+public struct AsyncWithLatestFrom2Sequence<Base: AsyncSequence, Other1: AsyncSequence, Other2: AsyncSequence>: AsyncSequence
+where Other1: Sendable, Other2: Sendable, Other1.Element: Sendable, Other2.Element: Sendable {
+  public typealias Element = (Base.Element, Other1.Element, Other2.Element)
+  public typealias AsyncIterator = Iterator
+
+  let base: Base
+  let other1: Other1
+  let other2: Other2
+
+  // for testability purpose
+  var onBaseElement: (@Sendable (Base.Element) -> Void)?
+  var onOther1Element: (@Sendable (Other1.Element?) -> Void)?
+  var onOther2Element: (@Sendable (Other2.Element?) -> Void)?
+
+  init(_ base: Base, _ other1: Other1, _ other2: Other2) {
+    self.base = base
+    self.other1 = other1
+    self.other2 = other2
+  }
+
+  public func makeAsyncIterator() -> Iterator {
+    var iterator = Iterator(
+      base: self.base.makeAsyncIterator(),
+      other1: self.other1,
+      other2: self.other2
+    )
+    iterator.onBaseElement = onBaseElement
+    iterator.onOther1Element = onOther1Element
+    iterator.onOther2Element = onOther2Element
+    iterator.startOthers()
+    return iterator
+  }
+
+  public struct Iterator: AsyncIteratorProtocol {
+    enum Other1State {
+      case idle
+      case element(Result<Other1.Element, Error>)
+    }
+
+    enum Other2State {
+      case idle
+      case element(Result<Other2.Element, Error>)
+    }
+
+    struct OthersState {
+      var other1State: Other1State
+      var other2State: Other2State
+
+      static var idle: OthersState {
+        OthersState(other1State: .idle, other2State: .idle)
+      }
+    }
+
+    enum BaseDecision {
+      case pass
+      case returnElement(Result<Element, Error>)
+    }
+
+    var base: Base.AsyncIterator
+    let other1: Other1
+    let other2: Other2
+
+    let othersState: ManagedCriticalState<OthersState>
+    var othersTask: Task<Void, Never>?
+
+    var isTerminated: ManagedCriticalState<Bool>
+
+    // for testability purpose
+    var onBaseElement: (@Sendable (Base.Element) -> Void)?
+    var onOther1Element: (@Sendable (Other1.Element?) -> Void)?
+    var onOther2Element: (@Sendable (Other2.Element?) -> Void)?
+
+    public init(base: Base.AsyncIterator, other1: Other1, other2: Other2) {
+      self.base = base
+      self.other1 = other1
+      self.other2 = other2
+      self.othersState = ManagedCriticalState(.idle)
+      self.isTerminated = ManagedCriticalState(false)
+    }
+
+    mutating func startOthers() {
+      self.othersTask = Task { [othersState, other1, other2, onOther1Element, onOther2Element] in
+        await withTaskGroup(of: Void.self) { group in
+          group.addTask {
+            do {
+              for try await element in other1 {
+                othersState.withCriticalRegion { state in
+                  state.other1State = .element(.success(element))
+                }
+                onOther1Element?(element)
+              }
+            } catch {
+              othersState.withCriticalRegion { state in
+                state.other1State = .element(.failure(error))
+              }
+            }
+          }
+
+          group.addTask {
+            do {
+              for try await element in other2 {
+                othersState.withCriticalRegion { state in
+                  state.other2State = .element(.success(element))
+                }
+                onOther2Element?(element)
+              }
+            } catch {
+              othersState.withCriticalRegion { state in
+                state.other2State = .element(.failure(error))
+              }
+            }
+          }
+        }
+      }
+    }
+
+    public mutating func next() async rethrows -> Element? {
+      let shouldReturnNil = self.isTerminated.withCriticalRegion { $0 }
+      guard !shouldReturnNil else { return nil }
+
+      return try await withTaskCancellationHandler { [isTerminated, othersTask] in
+        isTerminated.withCriticalRegion { isTerminated in
+          isTerminated = true
+        }
+        othersTask?.cancel()
+      } operation: { [othersTask, othersState, onBaseElement] in
+        do {
+          while true {
+            guard let baseElement = try await self.base.next() else {
+              isTerminated.withCriticalRegion { isTerminated in
+                isTerminated = true
+              }
+              othersTask?.cancel()
+              return nil
+            }
+
+            onBaseElement?(baseElement)
+
+            let decision = othersState.withCriticalRegion { state -> BaseDecision in
+              switch (state.other1State, state.other2State) {
+                case (.idle, _):
+                  return .pass
+                case (_, .idle):
+                  return .pass
+                case (.element(.success(let other1Element)), .element(.success(let other2Element))):
+                  return .returnElement(.success((baseElement, other1Element, other2Element)))
+                case (.element(.failure(let otherError)), _):
+                  return .returnElement(.failure(otherError))
+                case (_, .element(.failure(let otherError))):
+                  return .returnElement(.failure(otherError))
+              }
+            }
+
+            switch decision {
+              case .pass:
+                continue
+              case .returnElement(let result):
+                return try result._rethrowGet()
+            }
+          }
+        } catch {
+          isTerminated.withCriticalRegion { isTerminated in
+            isTerminated = true
+          }
+          othersTask?.cancel()
+          throw error
+        }
+      }
+    }
+  }
+}
+
+extension AsyncWithLatestFrom2Sequence: Sendable where Base: Sendable { }
--- a/Sources/AsyncAlgorithms/AsyncWithLatestFromSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncWithLatestFromSequence.swift
@@ -1,0 +1,169 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+public extension AsyncSequence {
+  /// Combines `self` with another ``AsyncSequence`` into a single ``AsyncSequence`` where each
+  /// element from `self` is aggregated to the latest known element from the `other` sequence (if any) as a tuple.
+  ///
+  /// Remark: as the `other` sequence is being iterated over in the context of its own ``Task``, there is no guarantee
+  /// that its latest know element is the one that has just been produced when the base sequence produces its next element.
+  ///
+  /// ```
+  /// let base = AsyncChannel<Int>()
+  /// let other = AsyncChannel<String>()
+  /// let sequence = base.withLatest(from: other)
+  ///
+  /// Task {
+  ///   for element in await sequence {
+  ///    print(element)
+  ///   }
+  /// }
+  ///
+  /// await other.send("a")
+  /// await other.send("b")
+  ///
+  /// ... later in the application flow
+  ///
+  /// await base.send(1)
+  ///
+  /// // will print: (1, "b")
+  /// ```
+  ///
+  /// - Parameter other: the other ``AsyncSequence``
+  /// - Returns: an ``AsyncWithLatestFromSequence`` where elements are a tuple of an element from `self` and the
+  /// latest known element (if any) from the `other` sequence.
+  func withLatest<Other: AsyncSequence>(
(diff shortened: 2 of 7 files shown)


Sources/AsyncAlgorithms/Rethrow.swift
```diff
@@ -26,7 +26,7 @@ extension _ErrorMechanism {
     fatalError("materialized error without being in a throwing context")
   }
   
-  internal func _rethrowGet() rethrows -> Output {
+  @usableFromInline internal func _rethrowGet() rethrows -> Output {
```
phausler, line 29: we should avoid this; the `_ErrorMechanism` is a work-around for a compiler feature that is not yet there so ideally it should not be exposed into ABI.

Tests/AsyncAlgorithmsTests/TestZipLatestFrom.swift
```diff
@@ -0,0 +1,782 @@
...
+    await base.finish()
+
+    wait(for: [finished], timeout: 1.0)
+    value = validator.current
+    XCTAssertEqual(value, [(1, "a"), (2, "a"), (3, "c")])
+  }
+
+  func test_zipLatestFrom_uses_latest_element_from_other_when_base_produces_first_elements() async {
+    // Timeline
+    // base:     |-1-2---3------4---------5-----
```
phausler, line 183: these diagrams are really close to the validation diagram, however the use of `|` is kinda confusing.

Perhaps we can use some sort of combing function that is easily understandable to represent the tuple? e.g. we can use emoji so a tuple of a shape out of 🤍⬜️⚪️ along side a color out of 🟥🟧🟨🟩🟦🟪⬛️ results in a symbol of ❤️🟥🔴🧡🟧🟠... etc?

Doing a 3 variant might be hard; I still need to think on how to accomplish that.

Sources/AsyncAlgorithms/AsyncZipLatestFromSequence.swift
```diff
@@ -0,0 +1,382 @@
...
+
+      var hasElement = false
+      repeat {
+        do {
+          baseElement = try await self.base.next()
+        } catch {
+          self.other.task.cancel()
+          throw error
+        }
+        lastKnownOtherState = await self.other.next()
```
phausler, line 157: so this does not run the two sequences concurrently for iteration. effectively it is blocking iteration from one to the other. Perhaps a TaskGroup is in order here (like how zip does it)?

Sources/AsyncAlgorithms/AsyncZipLatestFromSequence.swift
```diff
@@ -0,0 +1,382 @@
...
+  ///
+  /// await base.send(1)
+  ///
+  /// // will print: (1, "b")
+  /// ```
+  ///
+  /// - Parameter other: the other ``AsyncSequence``
+  /// - Returns: an ``AsyncZipLatestFromSequence`` where elements are a tuple of an element from `self` and the
+  /// latest known element (if any) from the `other` sequence.
+  @inlinable
```
phausler, line 39: I have a feeling that these probably should not be inlinable. Usually the bar I used was "could this reasonably be used for byte streams?" If the answer was yes then it should be inlined/frozen etc. If the answer is meh that doesn't make any sense then they are ok to be resilient boundaries. If the answer is on the fence then it is done if and only if it can be done w/o exposing hacks/workarounds for the language etc.

Tests/AsyncAlgorithmsTests/TestZipLatestFrom.swift
```diff
@@ -0,0 +1,782 @@
+//
+//  TestZipLatestFrom.swift
+//  
+//
+//  Created by Thibault Wittemberg on 01/04/2022.
+//
+
+@preconcurrency import AsyncAlgorithms
```
phausler, line 8: oof wat?
XCTest I get, but the framework we are building for async/await stuff; im sure this is just a copy/paste error?

Sources/AsyncAlgorithms/AsyncZipLatestFromSequence.swift
```diff
@@ -0,0 +1,382 @@
...
+    from other1: Other1,
+    _ other2: Other2
+  ) -> AsyncZipLatestFrom2Sequence<Self, Other1, Other2> {
+    AsyncZipLatestFrom2Sequence(self, other1, other2)
+  }
+}
+
+/// ``AsyncZipLatestFromSequence`` is an ``AsyncSequence`` where elements are a tuple of an element from `base` and the
+/// latest known element (if any) from the `other` sequence.
+public struct AsyncZipLatestFromSequence<Base: AsyncSequence, Other: AsyncSequence>: AsyncSequence where Other: Sendable {
```
phausler, line 93: these types probably need to be audited for being sendable. conditionally or not; my guess is that they will have similar constraints to zip/combineLatest.

Tests/AsyncAlgorithmsTests/TestZipLatestFrom.swift
```diff
@@ -0,0 +1,782 @@
...
+      self.lastKnownElement = baseElement
+      return baseElement
+    }
+  }
+}
+
+final class TestZipLatestFrom: XCTestCase {}
+
+// MARK: test for AsyncZipLatestSequence
+extension TestZipLatestFrom {
```
phausler, line 55: it would be good to get some coverage numbers so that we can get an idea of how much is addressed in the implementation by these tests; generally we aimed to have 90%+ coverage. Obviously coverage does not mean we cover ALL permutations of usage but it definitely assuages stability concerns etc.

Sources/AsyncAlgorithms/AsyncZipLatestFromSequence.swift
```diff
@@ -0,0 +1,382 @@
...
+
+    @usableFromInline
+    init(_ base: Base.AsyncIterator, _ other: AsyncCurrentElementSequence<Other>.AsyncIterator) {
+      self.base = base
+      self.other = other
+    }
+
+    @inlinable
+    public mutating func next() async rethrows -> Element? {
+      guard !Task.isCancelled else {
```
phausler, line 140: so the cancellation only happens at the start of the iteration? it does not respond immediately while the base here is going too iterated (or when the other is iterated?

phausler: If I am reading it correctly; I _think_ that is fine... just wanted to make sure that was the intent.

phausler (commented):
This looks like a fantastic start!

Sources/AsyncAlgorithms/Rethrow.swift
twittemb: removed :-)

Sources/AsyncAlgorithms/AsyncZipLatestFromSequence.swift
twittemb: removed :-)

Tests/AsyncAlgorithmsTests/TestZipLatestFrom.swift
twittemb: removed :-) It was due to a lack of conformance to Sendable

Tests/AsyncAlgorithmsTests/TestZipLatestFrom.swift
twittemb: I'm having some trouble getting the coverage in Xcode because of an error "Failed to merge raw profiles ...". Does that ring a bell?

Sources/AsyncAlgorithms/AsyncZipLatestFromSequence.swift
twittemb: I've reworked the cancellation handling so that every iteration can catch it

Sources/AsyncAlgorithms/AsyncZipLatestFromSequence.swift
twittemb: The `Other` sequence is being iterated over in its own Task. The call to `self.other.next()` will return a value right away (either noElement, element(value) or nil). With the way it is implemented right now I don't see the benefit of having a TaskGroup.

However, the current implementation might not be the good one! What do you think of iterating over `Other` in a separate Task so that we can always have its latest known value at our disposal ? It comes with the drawback that the latest known value might not be the one that we think because the Task's execution is scheduled by the system and it might be paused for some reason.

parkera:
Marking this as v1.1 because we should strongly consider it after our initial API stable 1.0 release.

Sources/AsyncAlgorithms/AsyncWithLatestFrom2Sequence.swift
```diff
@@ -0,0 +1,261 @@
...
+          other2Task?.cancel()
+          throw error
+        }
+      }
+    }
+  }
+}
+
+extension AsyncWithLatestFrom2Sequence: Sendable where Base: Sendable, Other1: Sendable, Other2: Sendable { }
+extension AsyncWithLatestFrom2Sequence.Iterator: Sendable where Base.AsyncIterator: Sendable, Other1: Sendable, Other2: Sendable { }
```
FranzBusch, line 261: The `Iterator` really shouldn't be `Sendable` here. The implementation also doesn't need it as far as I can see.

Sources/AsyncAlgorithms/AsyncWithLatestFrom2Sequence.swift
```diff
@@ -0,0 +1,261 @@
...
+  let other1: Other1
+  let other2: Other2
+
+  init(_ base: Base, _ other1: Other1, _ other2: Other2) {
+    self.base = base
+    self.other1 = other1
+    self.other2 = other2
+  }
+
+  public func makeAsyncIterator() -> Iterator {
```
FranzBusch, line 73: We should make sure that only a single Iterator can be created. The problem with multiple iterators is that it leads to non-deterministic behaviour where elements are ending up in.

Sources/AsyncAlgorithms/AsyncWithLatestFrom2Sequence.swift
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    var base: Base.AsyncIterator
+    let other1: Other1
+    let other2: Other2
+
+    let othersState: ManagedCriticalState<OthersState>
+
+    var other1Task: Task<Void, Never>?
+    var other2Task: Task<Void, Never>?
```
FranzBusch, line 115: Instead of spawning two tasks can we spawn one and two child tasks?

Sources/AsyncAlgorithms/AsyncWithLatestFromSequence.swift
```diff
@@ -0,0 +1,184 @@
...
+
+    var base: Base.AsyncIterator
+    let other: Other
+    let otherState: ManagedCriticalState<OtherState>
+    var otherTask: Task<Void, Never>?
+    var isTerminated: ManagedCriticalState<Bool>
+
+    // for testability purpose
+    var onBaseElement: (@Sendable (Base.Element) -> Void)?
+    var onOtherElement: (@Sendable (Other.Element?) -> Void)?
```
FranzBusch, line 91: I strongly believe we shouldn't insert code into these things just for testing. The behaviour should be fully observable from the public APIs.

Sources/AsyncAlgorithms/AsyncWithLatestFrom2Sequence.swift
```diff
@@ -0,0 +1,261 @@
...
+
+    var isTerminated: ManagedCriticalState<Bool>
+
+    // for testability purpose
+    var onBaseElement: (@Sendable (Base.Element) -> Void)?
+    var onOther1Element: (@Sendable (Other1.Element?) -> Void)?
+    var onOther2Element: (@Sendable (Other2.Element?) -> Void)?
```
FranzBusch, line 122: Same here

Sources/AsyncAlgorithms/AsyncWithLatestFromSequence.swift
```diff
@@ -0,0 +1,184 @@
...
+          otherTask?.cancel()
+          throw error
+        }
+      }
+    }
+  }
+}
+
+extension AsyncWithLatestFromSequence: Sendable where Base: Sendable, Other: Sendable { }
+extension AsyncWithLatestFromSequence.Iterator: Sendable where Base.AsyncIterator: Sendable, Other: Sendable { }
```
FranzBusch, line 184: Same here

Sources/AsyncAlgorithms/AsyncWithLatestFrom2Sequence.swift
```diff
@@ -0,0 +1,261 @@
...
+    mutating func startOthers() {
+      self.startOther1()
+      self.startOther2()
+    }
+
+    public mutating func next() async rethrows -> Element? {
+      let shouldReturnNil = self.isTerminated.withCriticalRegion { $0 }
+      guard !shouldReturnNil else { return nil }
+
+      self.startOthers()
```
FranzBusch, line 199: I think we need to start the `Task` in the `makeAsyncIterator` method to kick-off the subscription on the other sequences as fast as possible. This allows the first element from the base to have a higher chance to already have an element from the other sequence

FranzBusch (commented):
Just left some smaller comments here already that I spotted

FranzBusch:
> Hi @FranzBusch thanks for the review. I'll address them soon. As for the "ad-hoc" internal testing properties, I followed some advices given in this thread by Joe Groff (https://forums.swift.org/t/reliably-testing-code-that-adopts-swift-concurrency/57304/32). Testing concurrently executing code involving a time line is really challenging. In the case of `withLatest(from:)` we have to have a precise coordination that is deterministic because we have no control on the inner task execution for the other sequence.
> 
> What would you suggest to improve that ?


I am really torn on that suggestion, but I feel your pain points because I experienced the same in the `merge` and `debounce` implementation where certain scenarios can't be tested since you are not controlling the flow in the spawned Tasks. I think if we go forward with hooks for tests we should `#if DEBUG` to only have them in the tests but even that produces problems since we want to run the benchmarking tests in release mode.

I would like to get @phausler opinion here about test hooks for simulating timings.

Sources/AsyncAlgorithms/AsyncWithLatestFrom2Sequence.swift
twittemb: Can you be a bit more specific, I don't understand what you are asking for ?

FranzBusch: Sure, the problem here is a bit broader to the `AsyncSequence` ecosystem and I am planning to write something up soon. Right now in all the `AsyncSequence` implementations the `makeAsyncIterator` method can be called multiple times. Each of these iterators is almost always using the same underlying storage. So the actual implementation is that the algorithms are unicast `AsyncSequence`s however since they allow multiple iterators are getting into an awkward state where multiple `Task`s can consume them now. This causes all kinds of issues like non-deterministic behaviour who gets an element. Furthermore, cancellation handling is becoming problematic since if two `Task`s are consuming the same `AsyncSequence` and one gets cancelled it almost always cancels the other consumer as well. I currently consider it best practice to just guard that only a single Iterator can be created and `fatalError` on subsequent calls.

The current behaviour is fine for `Sequence`s since the they don't have this time aspect to them and are most of the time referencing an underlying content. For our `AsyncSequence`s the content is almost never buffered but consumed on the time stream.

As I said, I am planning to write more up on this soonish just wanted to make sure to put this out here already.

twittemb: Thanks for the detailed explanation. I'm not sure I see a problem since a call to `makeAsyncIterator()` creates a new iterator that has its own properties and storage. Nothing is shared between several iterators.

Am I missing something ?

For AsyncSequences that have a storage shared across consumers (like AsyncChannel or an hypothetical AsyncShareSequence),  the locking mechanism enforces the good behaviour right ?

FranzBusch: Ah sorry I missed that the whole core implementation is inside the Iterator. While this makes things a bit better since the state doesn't interfere it makes still for some weird behaviour IMO:

```swift
let other = AsyncChannel<>
let base = AsyncStream<>

let sequence = base.withLatest(from: other)

Task {
    for await element in sequence {}
}


Task {
    for await element in sequence {}
}
```

You cannot tell from this code what is going to happen since the way all of the sequences involved in this little snippet right now allow creating multiple iterators. While your implementation is multi-cast all the other ones are not. IMO, what I personally think we should do is implement almost all our algorithms and base sequences as unicast (unless they have different semantics) and provide `share/multicast` algorithms. This makes reasoning about the code in the end way easier

```swift
let other = AsyncChannel<>
let base = AsyncStream<>

let sequence = base.shared().withLatest(from: other.shared())
```

I don't think you have to change your PR here right away since this is still in discussion and like I said I wanna open up a broader discussion around this soonish. Just wanted to share more details with you in why I think we need to be careful around these things.

twittemb: Ok I understand. This is something I struggled with also. But it is mainly due to the fact that although AsyncChannel and AsyncStream are compatible with a multi producer/multi consumer paradigm, their output is not shared and the elements will be distributed across the consumers.

I guess this is what channels are for. If I refer to the Kotlin implementation it does exactly that. 

I've open sourced a project `AsyncExtensions` before `swift-async-algorithms` was announced :-) and I'm trying to address notably the share operator in there, as well as shared "channels". Now that `swift-async-algorithms` has been announced, I'm rethinking my repo as a companion to `swift-async-algorithms`. It means I will add things that don't yet exist in `swift-async-algorithms` and try to suggest them to this main repo step by step.

The repo is a work in progress (https://github.com/sideeffect-io/AsyncExtensions/tree/refactor) and the README is not up-to-date but there are some interesting operators in there, that were challenging, like `.switchToLatest()`.

FranzBusch: I think the `AsyncChannel` being MPMC makes sense; however, I am still very torn if `AsyncStream` should be like that. `AsyncChannel` does handle cancellation better since it only removes the continuations for that specific `Task` whereas `AsyncStream` cancels the whole thing which makes it very hard to use safely.

But we are learning all of this along the way so we should just come up with some rules around what to expect!

W.r.t. your package. I really like it! I also have a rudimentary implementation of `share` lying around; however, as @phausler said we first need to stabilise the algos that are in here and then add more.

Sources/AsyncAlgorithms/AsyncWithLatestFromSequence.swift
twittemb: @phausler do you have an opinion about that ?

Sources/AsyncAlgorithms/AsyncWithLatestFrom2Sequence.swift
twittemb: BTW, speaking of AsyncChannel cancellation, I have on ongoing PR regarding cancellation when several producers in different tasks -> https://github.com/apple/swift-async-algorithms/pull/184

Sources/AsyncAlgorithms/AsyncWithLatestFromSequence.swift
phausler: the more exposed APIs we have the more potential bugs/bad-interactions there could be. If we need some sort of mechanism for testing then we should figure out a way to observe the issue via the tests. Since the tests can control the stepping of the global executors, and we also control both the inputs AND the demand, we should be able to observe everything that is of importance in the tests.

phausler:
So we can control the spawning of tasks! https://github.com/apple/swift-async-algorithms/blob/main/Sources/AsyncSequenceValidation/Test.swift#L313 is where the tests do it today - and I could imagine that we could add either extra functionality to that to adjust the way we need... or we can use that dark magic to build some other really neat tools to control that in different ways.

What I think would be a good plan is to favor testing via what is public surface area as much as possible and then collect info on exactly what we need such that either a) we can get effort from a language/runtime layer or b) (ab)use some of the runtime functions to manipulate the execution into what we want to test.

Keithbreadley (approved):

phausler:
@twittemb I know it has been a while since we have looked at this; but this is one of the higher requested algorithms folks look for - if you would be interested in taking this back up, it is something I would like to consider here soon. If not, I would be willing to take this forward with modification (and of course give credit where it is due).

twittemb:
> @twittemb I know it has been a while since we have looked at this; but this is one of the higher requested algorithms folks look for - if you would be interested in taking this back up, it is something I would like to consider here soon. If not, I would be willing to take this forward with modification (and of course give credit where it is due).

Hi @phausler. I think you are best suited if we want to move forward on this topic since I’ve kinda lost my focus on this matter lately (I’m now a professional bladesmith in France 😜). I hope my original implementation will help I thank you in advance for your work. Have a nice one.

Thibault.

mini-min:
Hi @phausler, any updates on this one? 
Asking because I've needed this operator a few times myself and ended up hand-rolling it, so I'd be happy to pick this up if nobody is working on it.

Since the branch is pretty far behind main at this point (conflicts, plus the state machine rewrite happened after this), I think it would have to be a new PR anyway, keeping @twittemb's work credited. Let me know if that's ok with you.

Issue #148, opened 2022-04-15
More detailed testing needs to be done tracking the transitions for AsyncChannel


phausler:
The changes associated with https://github.com/apple/swift-async-algorithms/commit/047ab6abef71bdb6758c7bed661a04c04f5b056a need to have some sort of test that triggers that previous behavior to prove that the change listed definitely does what we think it does (which I am pretty sure it is correct). Having tests to prove its correctness will aide in preventing possible regressions etc.

twittemb:
Just for the record: both `AsyncChannel` and `AsyncThrowingChannel` should have those tests.

FranzBusch:
We can close this issue since we have re-implemted the AsyncChannel and added tests.

PR #150, opened 2022-04-20
asyncThrowingChannel: fixes potential crashes when no more awaiting

twittemb:
This PR is the sister from this one https://github.com/apple/swift-async-algorithms/pull/143 but for `AsyncThrowingChannel` that suffers from the very same issue.

The issue is described here: https://github.com/apple/swift-async-algorithms/issues/142

Before this PR, an `AsyncThrowingChannel` could crash in the following context:

- start an iteration in a Task
- cancel the task while the `AsyncThrowingChannel` is being awaited
- send a value in the channel

An issue has already been created to track the implementation of unit tests to ensure non regression: https://github.com/apple/swift-async-algorithms/issues/148


phausler (approved):
Yea, same fix, same previous caveat that we need to follow up with some way of validating this in tests; looks right to me.

Issue #136, opened 2022-04-02
Does `AsyncChannel` supports multiple producers/consumers?


laclouis5:
I was playing with `AsyncChannel` and tried to implement a multiple producers multiple consumers pattern but stumbled on an issue: I wasn't able to terminate the async iteration of all consumers. When there is one consumer the solution is simple, just call `AsyncChannel.finish()` at the end of the production to send a `nil` value to the only consumer which then stops.

However, when there are multiple consumers all of them should receive a `nil` value from the producers. To achieve this I tried to call `.finish()` as many time as there are consumers but only one stopped.

Is this intended behavior, i.e. an `AsyncChannel` should only be used with one consumer, or is this a bug? Please see the code snippet bellow to reproduce it.

System:
MacBook Air early 2015
macOS v12.3.1
Xcode v13.3
Swift v5.6 

PS: I just read in the documentation that `.finish()` returns immediately if the channel is already finished, so this means that calling it multiple times won't work for multiple consumers.

```swift
import Foundation
import AsyncAlgorithms

func producer(channel: AsyncChannel<Int>, id: Int) async {
    for _ in 0..<3 {
        await channel.send(id)
    }
}

/// Spawns `count` producers
func produce(channel: AsyncChannel<Int>, count: Int) async {
    await withTaskGroup(of: Void.self) { group in
        for id in 0..<count {
            group.addTask {
                await producer(channel: channel, id: id)
            }
        }
    }
    
    for _ in 0..<count {
        await channel.finish()
    }
}

func consumer(channel: AsyncChannel<Int>, id: Int) async {
    for await value in channel {
        try? await Task.sleep(nanoseconds: 1_000_000_000)
        print("Consumer \(id) consumed value produce by producer \(value)")
    }
    print("Consumer \(id) terminated")  // Only one consumer is stopped
}

/// Spawns `count` consumers
func consume(channel: AsyncChannel<Int>, count: Int) async {
    await withTaskGroup(of: Void.self) { group in
        for id in 0..<count {
            group.addTask {
                await consumer(channel: channel, id: id)
            }
        }
    }
}

@main
enum App {
    static func main() async throws {
        let channel = AsyncChannel<Int>()
        
        async let p: Void = produce(channel: channel, count: 4)
        async let c: Void = consume(channel: channel, count: 4)
        
        _ = await (p, c)
    }
}
```

phausler:
I think this is perhaps a bug; finish should terminate everything awaiting it

[...]

PR #152, opened 2022-05-01
Fix async channel finishes all

twittemb:
This PR aims to fix/improve the "finish" strategy for `AsyncChannel`/`AsyncThrowingChannel`.

From my understanding there are two possible scenarios when sending a finish:

**- 1:** There are no awaiting clients. There are potential pending "send" operations. We send finish 1 or several times.
**- 2:** There are no pending operations. There are several awaiting clients. We send finish 1 or several times.

The current behaviour is:

**- 1:** The state is set to `terminal`. Finish is added as a pending operation (other future finishes or values will be discarded). The future clients will consume the pending operations until they reach the finish operation. Once a client has consumed this finish operation, the next potential awaiting clients will be suspended for ever.
**- 2:** The state is set to `terminal`. The first awaiting operation is resumed with the nil value. The potential other awaiting operations won't be resumed. There is also infinite suspensions for them.

The behaviour with this PR is: 

**- 1 and 2:** All pending and awaiting operations are resumed with nil (other furure finishes or values will be discarded). The potential futur clients will be instantly resumed with nil.

Some considerations that might trigger questions:
- the `finish` function is no more async since it does not involve the creation of a continuation.
- If there are several pending operations when finish is called, they will be discarded, meaning that if we have several Tasks that send values in a `Channel` and one of them call `finish` then some potential suspended `send` in other tasks will be resumed instantly and their value will never be emitted.
- Unit testing the scenario where `finish` is called while several clients are already awaiting is difficult because there is no way to wait for the `next` function to be called before calling `finish`. Ideas are welcome.

This PR addresses the issue described here -> https://github.com/apple/swift-async-algorithms/issues/136

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncChannel.swift
+++ b/Sources/AsyncAlgorithms/AsyncChannel.swift
@@ -131,7 +131,12 @@ public final class AsyncChannel<Element: Sendable>: AsyncSequence, Sendable {
   func next(_ generation: Int) async -> Element? {
     return await withUnsafeContinuation { continuation in
       var cancelled = false
+      var terminal = false
       state.withCriticalRegion { state -> UnsafeResumption<UnsafeContinuation<Element?, Never>?, Never>? in
+        if state.terminal {
+          terminal = true
+          return nil
+        }
         switch state.emission {
         case .idle:
           state.emission = .awaiting([Awaiting(generation: generation, continuation: continuation)])
@@ -157,13 +162,13 @@ public final class AsyncChannel<Element: Sendable>: AsyncSequence, Sendable {
           return nil
         }
       }?.resume()
-      if cancelled {
+      if cancelled || terminal {
         continuation.resume(returning: nil)
       }
     }
   }
   
-  func cancelSend() {
+  func finishAll() {
     let (sends, nexts) = state.withCriticalRegion { state -> ([UnsafeContinuation<UnsafeContinuation<Element?, Never>?, Never>], Set<Awaiting>) in
       if state.terminal {
         return ([], [])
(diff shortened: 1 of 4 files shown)


Sources/AsyncAlgorithms/AsyncChannel.swift
```diff
@@ -224,20 +221,19 @@ public final class AsyncChannel<Element: Sendable>: AsyncSequence, Sendable {
...
-      await _send(.success(element))
+    await _send(element)
   }
   
-  /// Send a finish to an awaiting iteration. This function will resume when the next call to `next()` is made.
-  /// If the channel is already finished then this returns immediately
-  public func finish() async {
-      await _send(.success(nil))
+  /// Send a finish to all awaiting iterations.
+  public func finish() {
```
phausler, line 235: making this non async might be a problem from an isolation standpoint. I think that perhaps there is now a scenario where the iteration ends up hanging because a pending nil is awaiting consumption.

twittemb: Could you give an example so I can picture what you have in mind ?

phausler: so I went over the implementations; this seems like we are in the clear for this.

twittemb: Ok great news. Thanks for the follow up.

twittemb:
> Couple of tasks to wrap this up:
> 
> * [ ]  Update the documentation for finish to reflect it's non-awaiting characteristics.
> * [ ]  Update the guide to reflect the historical designs here and the new non-async-ness of the finish function

Definitely. I will do that next week as I don’t have my mac with me this week.

parkera:
@swift-ci test

phausler:
the macOS tests fell over because it is executing the wrong test system

PR #153, opened 2022-05-03
Minor documentation improvements

ktan17:
Fixed a few grammatical/spacing typos and one missing line of sample code.


Sources/AsyncAlgorithms/AsyncChunkedByGroupSequence.swift
```diff
@@ -37,6 +37,7 @@ extension AsyncSequence {
 /// error, `AsyncChunkedByGroupSequence` will rethrow that error immediately and discard
 /// any current group.
 ///
+///      let numbers = [10, 20, 30, 10, 40, 40, 10, 20]
```
phausler, line 40: this should be an async sequence so tack on a `.async` on the end there

ktan17: oops, fixed!

Issue #139, opened 2022-04-07
Deadlock in Task.select when using clock's sleep 


mbrandonw:
I'm using Swift dev snapshot 2022-03-02 with Xcode 13.3.

By making a small tweak to [one](https://github.com/apple/swift-async-algorithms/blob/18749897a885e596d37efc1c4c911b4581d5b3cb/Tests/AsyncAlgorithmsTests/TestTaskSelect.swift#L17-L25) of the `Task.select` tests I am able to produce a deadlock:

```diff
 let firstValue = await Task.select(Task {
+  try! await Task.sleep(until: .now + .seconds(1), clock: .continuous)
   return 1
 }, Task {
   try! await Task.sleep(until: .now + .seconds(2), clock: .continuous)
   return 2
 }).value
```

It's worth noting that it does not deadlock if you use `Task.sleep` without a clock:

```swift
let firstValue = await Task.select(Task {
  try! await Task.sleep(nanoseconds: NSEC_PER_SEC)
  return 1
}, Task {
  try! await Task.sleep(nanoseconds: 2*NSEC_PER_SEC)
  return 2
}).value
```

Is there something about clocks and sleeping that is causing this? I see the same behavior with suspending clocks too.

kielgillard:
Is this PR related to this issue? https://github.com/apple/swift/pull/42527

mbrandonw:
@kielgillard, nice find! When I now run the test with the 2022-04-23 toolchain it no longer deadlocks, but the test does fail non-deterministically.

Also, regardless of the test passing or failing, it does so immediately (0.002 seconds runtime), and so clearly the continuous clock is not actually waiting for time to pass, which would explain the non-determinism:

```
Test Suite 'Selected tests' passed at 2022-04-25 08:23:12.181.
	 Executed 1 test, with 0 failures (0 unexpected) in 0.002 (0.003) seconds
```

On the other hand, the suspending clock is working exactly as I expect. The test passes 100% of the time (I just performed a repeating test with 100 iterations) and the test actually takes a second to execute:

```
Test Suite 'Selected tests' passed at 2022-04-25 08:24:20.392.
	 Executed 1 test, with 0 failures (0 unexpected) in 1.034 (1.040) seconds
```

@phausler Is there anything about `ContinuousClock` that would cause this behavior?

phausler:
yea the deadlock was not a deadlock but instead a stall out on certain time base calculations so yes it was related. 

[...]

PR #154, opened 2022-05-04
Test sleeping in both tasks of Task.select.

mbrandonw:
This makes it so that each task in `Task.select` sleeps so that we get test coverage on the behavior I saw in #139 where the task would hang forever. The task does not hang forever if one of the selections is a non-awaiting block of code.

We don't need to sleep for any significant amount of time. I chose 100 microseconds, but it could even be a 0 duration sleep.


phausler:
@swift-ci please test

phausler:
Task.select was removed since it caused design patterns to push to sendable iterators. So this is not needed anymore.

PR #156, opened 2022-05-12
[SAA-0002] Merge

phausler:
This is a modification of the merge guide for making it into a proposal.


twittemb:
If you agree with it, I’ll do a PR to align the unit test names with the ones from `zip`. I’ll do it afterwards since I don’t have access to my Mac for a few weeks.

phausler:
yea the test names can be updated w/o requirement to these proposals - if you want to tackle that broadly that is perfectly reasonable and appreciated contributions in my book and does not need any evolution review.

PR #159, opened 2022-06-10
Update availability to be in sync with Clock/Instant/Duration

phausler:
This annotates proper availability for the Clock/Instant/Duration usage.

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncChunksOfCountOrSignalSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncChunksOfCountOrSignalSequence.swift
@@ -31,25 +31,25 @@ extension AsyncSequence {
   }
 
   /// Creates an asynchronous sequence that creates chunks of a given `RangeReplaceableCollection` type of a given count or when an `AsyncTimerSequence` fires.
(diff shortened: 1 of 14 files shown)


phausler:
we should cut a tag for this after this is merged

parkera (approved):

rpassis:
H there,

I just noticed this was tagged as `0.0.2` even though there are breaking changes only available for Swift >= 5.7.

In looking at the README, this section:

> The Swift Async Algorithms package has a goal of being source stable as soon as possible; version numbers will follow [Semantic Versioning](https://semver.org/). Source breaking changes to public API can only land in a new major version.

Should we expect these breaking changes to continue to occur until the API reaches version 1.0.0? In other words should we be pinning this package to a fixed tag until 1.0.0 is released if we want to avoid breakages in our builds?

rpassis:
I'll answer my own question here, from https://semver.org/:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

parkera:
Regardless, we may still want to revert the tag as we think about if we want 0.0.x to have this change or if it goes into 1.0. @phausler has been looking into it.

PR #162, opened 2022-06-13
Compacted proposal

phausler:
This transforms the compacted guide into a proposal.


twittemb:
Shouldn't the evolution file be labeled SAA-0003 ?

phausler:
I have a few that are incoming this week of proposals; I will number them in the order they land.

twittemb (approved):
LGTM

PR #163, opened 2022-06-14
Improve Lock and ManagedCriticalState

Cyberbeni:
- make Lock private
- make Lock safe to use instead of requiring manually deallocating
- turn ManagedCriticalState into a class instead of a class wrapped in a struct

Haven't found any tests or comments about why all this stuff is currently implemented the way it is. I'm on my Windows PC, so couldn't check if it compiles.

### DIFF
--- a/Sources/AsyncAlgorithms/Locking.swift
+++ b/Sources/AsyncAlgorithms/Locking.swift
@@ -77,8 +77,8 @@ final class ManagedCriticalState<State> {
   }
   
   func withCriticalRegion<R>(_ critical: (inout State) throws -> R) rethrows -> R {
-    Lock.lock(lock)
-    defer { Lock.unlock(lock) }
+    Lock.lock(_lock)
+    defer { Lock.unlock(_lock) }
     return try critical(&_state)
   }
 }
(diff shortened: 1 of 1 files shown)


Sources/AsyncAlgorithms/Locking.swift
```diff
@@ -17,100 +17,69 @@
...
 
-struct ManagedCriticalState<State> {
-  private final class LockedBuffer: ManagedBuffer<State, Lock.Primitive> {
-    deinit {
-      withUnsafeMutablePointerToElements { Lock.deinitialize($0) }
-    }
-  }
-  
-  private let buffer: ManagedBuffer<State, Lock.Primitive>
+final class ManagedCriticalState<State> {
```
phausler, line 71: This type was structural because it was using a ManagedBuffer to store both the state and the lock as properly aligned and encapsulated memory. This approach wastes a secondary allocation and has the potential of loading the memory location of the state outside of critical regions guarded by the lock.

Sources/AsyncAlgorithms/Locking.swift
```diff
@@ -17,100 +17,69 @@
 @_implementationOnly import WinSDK
 #endif
 
-internal struct Lock {
+private final class Lock {
```
phausler, line 20: This is quite wasteful; it poses two allocations instead of just one location managed by an external item

phausler (commented):
Im not sure this is really an improvement to the lock, it uses potentially unsafe storage and ends up massively increasing the allocations associated with locking.

Issue #155, opened 2022-05-04
[Question `AsyncThrowingSequence`] Should failure resume and fail all producers/consumers


twittemb:
Hi,

We’ve recently merged a PR https://github.com/apple/swift-async-algorithms/pull/152 where the finish operation instantly resumes all the pending producers and consumers (with a nil value), thus ending all the iterators.

I was wondering if we should have the same behaviour for failures. As it is a terminal event, we have the same kind of risk for infinite suspensions. I guess when sending a failure, all the pending operations should resume instantly. All the awaiting consumers should receive the failure and the future iterations should also resume instantly.

@phausler what do you think ?

The `fail(_:)` function would also be  not async. It makes sense since, as a terminal event, no back pressure should be applied.

phausler:
the logic here is sound. The behavior we would ideally want for multiple consumers is that if a fail is sent, it needs to definitely be consumed; that is a long way of saying - I need to ruminate on this one a bit.

phausler:
so after thinking about the situations for this: provided it can be accomplished w/o starving either side (which I think I have convinced myself that is accomplishable as you outlined). From a symmetry standpoint it feels right to have back pressure be the same for finish and fail. In short that sounds like a good change to make.

twittemb:
Thanks for your analysis. I’ll work on that as soon as I can. I will keep the issue open until the PR is done so we can make the link between the issue and the PR.

twittemb:
https://github.com/apple/swift-async-algorithms/pull/164

PR #164, opened 2022-06-14
[AsyncThrowingChannel] make the `fail` terminal event non async

twittemb:
This PR is the sister of https://github.com/apple/swift-async-algorithms/pull/152 and is linked to that issue https://github.com/apple/swift-async-algorithms/issues/155.

Long story short: terminal events (finish/fail) should not imply a back pressure management since no further elements can be sent.

When fail is called:
- all pending and awaiting operations should be immediately resumed (in order to avoid potential infinite suspensions)
- further calls to `next()` will throw the error
- further calls to `send(_:)` will immediately resume

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncThrowingChannel.swift
+++ b/Sources/AsyncAlgorithms/AsyncThrowingChannel.swift
@@ -11,7 +11,13 @@
 
 /// An error-throwing channel for sending elements from on task to another with back pressure.
 ///
-/// The `AsyncThrowingChannel` class is intended to be used as a communication types between tasks, particularly when one task produces values and another task consumes those values. The back pressure applied by `send(_:)`, `fail(_:)` and `finish()` via suspension/resume ensures that the production of values does not exceed the consumption of values from iteration. Each of these methods suspends after enqueuing the event and is resumed when the next call to `next()` on the `Iterator` is made.
+/// The `AsyncThrowingChannel` class is intended to be used as a communication types between tasks,
+/// particularly when one task produces values and another task consumes those values. The back
+/// pressure applied by `send(_:)` via suspension/resume ensures that the production of values does
+/// not exceed the consumption of values from iteration. This method suspends after enqueuing the event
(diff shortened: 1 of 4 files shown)


phausler:
it might be worthwhile to investigate how this interacts with https://github.com/apple/swift-async-algorithms/pull/167

twittemb:
Hi. I think they collide. This PR also addresses the issue where the failure is lost. I encountered the problem while testing the “non async” version of the fail event.

Our implementation of the “Termination” enum is very similar expect i have decided to remove the “terminal” variable and to add a new “terminated” case to the the Emission enum. As “finish” or “fail” are terminal events, I think it makes sense to integrate this concept directly to the Emission state. This way we can’t be in both “pending” and “terminated” state at the same time.

What do you think ?

Sources/AsyncAlgorithms/AsyncThrowingChannel.swift
```diff
@@ -158,53 +171,77 @@ public final class AsyncThrowingChannel<Element: Sendable, Failure: Error>: Asyn
              state.emission = .awaiting(nexts)
            }
           return nil
+        case .terminated(let termination):
+          potentialTermination = termination
```
phausler, line 175: at this point the terminal state is consumed with the failure. So we wouldn't expect to hold onto that error indefinitely; instead it likely should follow the rules of AsyncSequence iteration and return nil for subsequent calls. To that end it probably should set the state.emission to `.terminated(.finished)`

phausler (commented):
Modulo the one issue with terminal transitions after it being emitted this looks great.

Sources/AsyncAlgorithms/AsyncThrowingChannel.swift
twittemb: good catch! I've adapted both the code and the unit test to reflect that.

twittemb:
> Modulo the one issue with terminal transitions after it being emitted this looks great.

thanks for the review @phausler. I've pushed a new version.

phausler (approved):
looks solid to me!

PR #165, opened 2022-06-15
Remove the disable of availability and mark methods as available to clock/instant/duration

phausler:
This annotates the tests and removes the unsafe flags to disable availability checking. Any test that touches clocks must be marked or gated for availability. Unsafe flags prevent other libraries from linking against this package; by removing this we now can support generally all potential linkages.

PR #167, opened 2022-06-16
Ensure AsyncThrowingChannel reliably throws errors when racing against iteration

phausler:
Another pull request (https://github.com/apple/swift-async-algorithms/pull/165) revealed a test that indicated flakiness in the potential terminal states of AsyncThrowingChannels when resolving to an error state. This ended up being that if the emission of the failure would happen before the iteration it would potentially set the state to finished, but not yield a failure, making subsequent calls to iteration produce nil.


phausler:
Closing in favor of pursuing this via https://github.com/apple/swift-async-algorithms/pull/164

PR #172, opened 2022-06-22
Fix typo in Validation.md

stzn:
I think we need to add escaping characters before ` |`.


phausler (commented):
image
That does not seem to fix the rendering, is this perhaps a GH bug?

phausler:
Perfect, that seems to have done the trick!

PR #177, opened 2022-07-13
Add an introduction and formatting for the `combineLatest` algorithm into a proposal.

phausler:
Read the full proposal [here](https://github.com/phausler/swift-async-algorithms/blob/pr/combine_latest_proposal/Evolution/NNNN-combineLatest.md).

### DIFF
--- a/Evolution/0006-combineLatest.md
+++ b/Evolution/0006-combineLatest.md
@@ -1,6 +1,6 @@
 # Combine Latest
 
-* Proposal: [SAA-NNNN](https://github.com/apple/swift-async-algorithms/blob/main/Evolution/NNNN-combineLatest.md)
+* Proposal: [SAA-0006](https://github.com/apple/swift-async-algorithms/blob/main/Evolution/0006-combineLatest.md)
 * Authors: [Philippe Hausler](https://github.com/phausler)
 * Status: **Implemented**
 
--- a/Evolution/NNNN-combineLatest.md
+++ b/Evolution/NNNN-combineLatest.md
@@ -1,0 +1,95 @@
+# Combine Latest
+
+* Proposal: [SAA-NNNN](https://github.com/apple/swift-async-algorithms/blob/main/Evolution/NNNN-combineLatest.md)
+* Authors: [Philippe Hausler](https://github.com/phausler)
+* Status: **Implemented**
+
+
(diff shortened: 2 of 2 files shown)


tcldr:
Excited to see this included. One comment:

The Naming part of the proposal mentions "withLatestFrom" as alternative naming. However, this is a quite different algorithm. Unlike, say CombineLatest2, which takes publisher A and publisher B, and emits when either A or B updates, `withLatestFrom` only emits when B updates. This can be useful for getting a snapshot of a frequently emitting sequence, say accelerometer data, on the occurrence of an infrequently emitting sequence, say a screen tap. e.g:
    for await (tap, orientation) in screenTaps.withLatestFrom(accelerometer) {
        // do something only on screen tap, with last emitted data from accelerometer
    }

Also,  what is the home for discussions on evolution proposals for Swift Algorithms? I couldn't find a CombineLatest thread in the swift forums and I don't see any discussions on the Issues board. The package readme mentions GitHub or Swift Forums, but isn't specific.

phausler:
I haven't posted them to the forums yet; was planning on doing that Monday. Per an additional algorithm - that is definitely something we should talk about but I want to get through the stuff we have first and then discuss the others with a bit more dedicated focus. But if you want to take that on we can park that as a pitch until then and get it queued up as an item to cover.

tcldr:
Lovely, makes sense. Thanks, Philippe.

twittemb:
> Excited to see this included. One comment:
> 
> The Naming part of the proposal mentions "withLatestFrom" as alternative naming. However, this is a quite different algorithm. Unlike, say CombineLatest2, which takes publisher A and publisher B, and emits when either A or B updates, `withLatestFrom` only emits when B updates. This can be useful for getting a snapshot of a frequently emitting sequence, say accelerometer data, on the occurrence of an infrequently emitting sequence, say a screen tap. e.g:
> 
> ```
> for await (tap, acceleration) in screenTaps.withLatestFrom(accelerometer) {
>     // do something only on screen tap, with last emitted data from accelerometer
> }
> ```
> 
> I'd love to see a proposal for this operator, too.
> 
> Also, what is the home for discussions on evolution proposals for Swift Algorithms? I couldn't find a CombineLatest thread in the swift forums and I don't see any discussions on the Issues board. The package readme mentions GitHub or Swift Forums, but isn't specific.

Hi,

If I may, there is already a proposal for `withLatestFrom` (temporarily named `zipLatestFrom`). The PR is [here](https://github.com/apple/swift-async-algorithms/pull/147) and the pitch is [here](https://forums.swift.org/t/pitch-withlatestfrom/56487/28)

twittemb:
Hi @phausler

From what I can see, on the one hand the implementation of `CombineLatest3` is just a shortcut on two `CombineLatest2`. On the other hand, the implementations of `Merge3` and `Zip3` are done from scratch. Why not also reusing  two `Merge2` and `Zip2` ? is there is performance overhead in doing this ?

Thanks.

PR #180, opened 2022-07-18
Proposal for `joined` and `joined(separator:)`

FranzBusch:
This adds a proposal for the `joined` and `joined(separator:)` algorithms

@phausler Could you check if there is anything missing. Also is it fine to have you as the author of the proposal since you implemented the algo?

Read the full proposal [here](https://github.com/apple/swift-async-algorithms/blob/ebfd880ca524ef09f4bc924464adcf593a8a05e4/Evolution/NNNN-joined.md)

### DIFF
--- a/Evolution/0004-joined.md
+++ b/Evolution/0004-joined.md
@@ -1,6 +1,6 @@
 # Joined
 
-* Proposal: [SAA-NNNN](https://github.com/apple/swift-async-algorithms/blob/main/Evolution/NNNN-joined.md)
+* Proposal: [SAA-0004](https://github.com/apple/swift-async-algorithms/blob/main/Evolution/0004-joined.md)
 * Authors: [Philippe Hausler](https://github.com/phausler)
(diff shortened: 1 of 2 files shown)


phausler:
That looks fine to me, if you want to put it on the forums and manage the thread associated with feedback for it please feel free to post.

twittemb:
Hi,

What is the difference between `joined` and `flatMap { $0 }` ?

kperryua:
> What is the difference between `joined` and `flatMap { $0 }` ?

They're semantically identical. However the **theory** is that knowing in advance that the incoming element is already an `AsyncSequence` and that no transformation will be happening on it will allow for a more optimized implementation.

It's worth noting that the stdlib also has both `joined()` and `flatMap()` on `Sequence`, although perhaps for different reasons than we do here. `Sequence.joined()` returns a `FlattenSequence` which lazily enumerates the constituent sequences, while `flatMap()` does so greedily.

Issue #182, opened 2022-07-22
[AsyncChannel] Question: is task cancellation in sending operations too violent?


twittemb:
Hi @phausler 

I was wondering if the task cancellation when sending an element in an AsyncChannel was not a bit too violent ?

This is the code when sending an element:

```
func _send(_ element: Element) async {
  await withTaskCancellationHandler {
    terminateAll()
  } operation: { ... }
```

the `terminateAll()` function will resume all the suspended operations (producers and consumers). What if the sending operations are performed in several Tasks? it means the async channel is finished for every one of them, which seem a bit violent right ?

twittemb:
I guess in case of a sending cancellation, we should only make sure the send continuation is resumed immediately. If we are in the case where there is an awaiting that is being resumed by the sending then we should also check for `Task.isCancelled` right before resuming the awaiting?

what to you think ?

twittemb:
https://github.com/apple/swift-async-algorithms/pull/184

twittemb:
PR is merged

PR #184, opened 2022-07-28
[Channel] improve send cancellation

twittemb:
This PR aims to:

- harmonize the cancellation handling for the `next()` and `send()` operations. Previously, a `send()` cancellation was terminating all the pending and awaiting operations, which seemed wrong to me as mentioned here https://github.com/apple/swift-async-algorithms/issues/182.
- improve the readability of the cancellation in the case it happens before the operation (send/next). Previously, a "cancelled" `next` was added to the awaiting list, so it can be immediately removed by the operation handling. Although it seemed to work well, it was not obvious when reading the code, and I tried to make it more explicit, so the maintenance is easier.


Sources/AsyncAlgorithms/AsyncChannel.swift
```diff
@@ -50,68 +53,49 @@ public final class AsyncChannel<Element: Sendable>: AsyncSequence, Sendable {
...
+
+  enum ChannelTokenStatus: Equatable {
+    case new
+    case cancelled
+  }
   
   enum Emission {
     case idle
-    case pending([UnsafeContinuation<UnsafeContinuation<Element?, Never>?, Never>])
+    case pending(Set<Pending>)
```
phausler, line 89: hmm making this a set becomes non-deterministic on the order of application - which will make it hard to test. Do we need set uniqueness to prevent issues or would just an Array work?

Sources/AsyncAlgorithms/AsyncChannel.swift
```diff
@@ -50,68 +53,49 @@ public final class AsyncChannel<Element: Sendable>: AsyncSequence, Sendable {
...
     case awaiting(Set<Awaiting>)
```
phausler, line 90: also the set of awaiting might pose issue too come to think of it... (I should have caught this earlier)

phausler (approved):

Sources/AsyncAlgorithms/AsyncChannel.swift
twittemb: Unfortunately we have to be able to uniquely identify the pending operations so we can remove/resume them in case of cancellation thanks to their "id".

Using an array as a FIFO could be nice to guarantee deterministic operation but we cannot rely on the index of a newly added operation to potentially remove it in case of a cancellation because the array could be updated by another concurrent send/next operation in the meantime. Also, if want want to keep an O(1) complexity for retrieving a pending operation, we need an indexed storage of some sort.

I guess we need something like OrderedSet but it would imply a dependency on swift-collections which we don't want I presume. Do you think we should implement a custom storage that fits our needs here ?

twittemb: @phausler i see in this PR https://github.com/apple/swift-async-algorithms/pull/185 that an dependency on Swift collections is introduced. Is this something we want ? We could leverage OrderedSet if so

twittemb: @phausler did you had the change to think a bit about having an external dep ?

phausler: OrderedSet is perfectly fine in my book.

twittemb: Nice, then I’ll update this PR to integrate it.
Thanks for your reply.

twittemb: @phausler I have successfully introduced OrderedSet in this PR

twittemb: @phausler do you think we have a go for this one, now that we use OrderedSet ?

phausler: yea I think we are good to go here

PR #185, opened 2022-08-10
Remove `AsyncIterator: Sendable` requirement from merge

FranzBusch:
# Motivation
Currently a lot of the operator implementations in here that consume other `AsyncSequence`s require the `AsyncIterator` to be `Sendable`. This is mostly due to the fact that we are calling `makeAsyncIterator` on the upstream `AsyncSequence` and then pass that iterator around to various newly spawned `Task`s. This has two downsides:
1. It only allows users to use operators like `merge` if their `AsyncSequence.AsyncIterator` is `Sendable`
2. In merge we are creating new `Task`s for every new demand. Creating `Task`s is not cheap.

My main goal of this PR was to remove the `Sendable` constraint from `merge`.

# Modification
This PR overhauls the complete inner workings of the `AsyncMerge2Sequence`. It does a couple of things:
1. The main change is that instead of creating new `Task`s for every demand, we are creating one `Task` when the `AsyncIterator` is created. This task has as child task for every upstream sequence.
2. When calling `next` we are signalling the child tasks to demand from the upstream
3. A new state machine that is synchronizing the various concurrent operations that can happen
4. Handling cancellation since we are creating a bunch of continuations.

# Result
In the end, this PR swaps the implementation of `AsyncMerge2Sequence` and drops the `Sendable` constraint and passes all tests. Furthermore, on my local performance testing I saw up 50% speed increase in throughput.

# Open points
1. I need to make this sequence re-throwing but before going down that rabbit whole I wanna get buy-in on the implementation.
2. We should discuss and document if `merge` and other operators are hot or cold, i.e. if they only request if they got downstream demand
3. I need to switch `AsyncMerge3Sequence` over to the same iplementation


Package.swift
```diff
@@ -16,9 +16,15 @@ let package = Package(
     .library(name: "_CAsyncSequenceValidationSupport", type: .static, targets: ["AsyncSequenceValidation"]),
     .library(name: "AsyncAlgorithms_XCTest", targets: ["AsyncAlgorithms_XCTest"]),
   ],
-  dependencies: [],
+  dependencies: [
+    .package(url: "https://github.com/apple/swift-collections.git", from: "1.0.2"),
```
FranzBusch, line 20: I was unsure if we should add this dependency. It makes the implementation a lot better and IMO is a lower level package. Open to discuss this :)

FranzBusch:
Switching this from `CheckedContinuation` to `UncheckedContinuation` is improving the performance by a lot:

Current main: 1x baseline
This PR with Checked: 1.5x
This PR with Unchecked: 3.5x

FranzBusch:
@phausler Yeah no rush on this! Enjoy your vacation and we can talk afterwards. I just wanted to put this up for discussion. I agree that re throwing is kinda coming the way but I am seeing up to 3.5 times performance improvements with this. However, performance is no the main driver, it really is the `Sendable` annotation. Let's discuss this once you are back!

Sources/AsyncAlgorithms/Merge2StateMachine.swift
```diff
@@ -0,0 +1,162 @@
...
+  var state: (PartialIteration<Base1.AsyncIterator, Partial>, PartialIteration<Base2.AsyncIterator, Partial>)
+
+  init(_ iterator1: Base1.AsyncIterator, terminatesOnNil iter1TerminatesOnNil: Bool = false, _ iterator2: Base2.AsyncIterator, terminatesOnNil iter2terminatesOnNil: Bool = false) {
+    self.iter1TerminatesOnNil = iter1TerminatesOnNil
+    self.iter2terminatesOnNil = iter2terminatesOnNil
+    state = (.idle(iterator1), .idle(iterator2))
+  }
+
+  mutating func apply(_ task1: Task<Merge2StateMachine<Base1, Base2>.Partial, Never>?, _ task2: Task<Merge2StateMachine<Base1, Base2>.Partial, Never>?) async rethrows -> Either? {
+    switch await Task.select([task1, task2].compactMap ({ $0 })).value {
```
phausler, line 38: do you think there is a way we can approach this without `Task.select`? That is the other perf bottle neck (which could be either solved by not needing it, or solved by improving it at a language level)

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
```diff
@@ -9,201 +9,1069 @@
...
+                                // is consuming the upstream sequence
+                                group.addTask {
+                                    var iterator1 = base1.makeAsyncIterator()
+
+                                    // This is our upstream consumption loop
+                                    loop: while true {
+                                        // We are creating a continuation before requesting the next
+                                        // element from upstream. This continuation is only resumed
+                                        // if the downstream consumer called `next` to signal his demand.
+                                        try await withCheckedThrowingContinuation { continuation in
```
phausler, line 166: since this is being validated by tests we can use the non checked variant

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
```diff
@@ -9,201 +9,1069 @@
...
+            )
+            /// Indicates that nothing should be done.
+            case none
+        }
+
+        mutating func iteratorDeinitialized() -> IteratorDeinitializedAction {
+            switch state {
+            case .initial:
+                // An iterator needs to be initialized before it can be deinitialized.
+                preconditionFailure("Internal inconsistency")
```
phausler, line 553: the preconditions in here make my spidey senses tingle.... any fatal error will become a bug (in my experience). is it possible to somehow refactor these to not be?

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
```diff
@@ -9,201 +9,1069 @@
...
+                    task: task,
+                    buffer: .init(),
+                    upstreamContinuations: [], // This should reserve capacity in the variadic generics case
+                    upstreamsFinished: 0,
+                    downstreamContinuation: nil
+                )
+
+            case .merging, .upstreamThrew, .finished:
+                // We only a single iterator to be created so this must never happen.
+                preconditionFailure("Invalid state")
```
phausler, line 602: it might be good to (if we decide these are useful to keep) list the state that is invalid and unify that with the same verbiage as the other fatal error cases.

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
```diff
@@ -9,201 +9,1069 @@
...
+
+            case .modifying:
+                preconditionFailure("Invalid state")
+            }
+        }
+
+        /// Actions returned by `childTaskSuspended()`.
+        enum ChildTaskSuspendedAction {
+            /// Indicates that the continuation should be resumed which will lead to calling `next` on the upstream.
+            case resumeContinuation(
```
phausler, line 612: stylistic nit; most folks have wide enough monitors to see the comment, so no need to limit code to 40 col... (cause if you are editing swift on a non emulated VT100... well time to upgrade :) )

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
```diff
@@ -9,201 +9,1069 @@
...
+        mutating func elementProduced(_ element: Element) -> ElementProducedAction {
+            switch state {
+            case .initial:
+                // Child tasks that are producing elements are only created after we transitioned to `merging`
+                preconditionFailure("Invalid state")
+
+            case let .merging(task, buffer, upstreamContinuations, upstreamsFinished, .some(downstreamContinuation)):
+                // We produced an element and have an outstanding downstream continuation
+                // this means we can go right ahead and resume the continuation with that element
+                precondition(buffer.isEmpty, "We are holding a continuation so the buffer must be empty")
```
phausler, line 694: this feels like it somehow could not be fully true... do we have tests that validate this?

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
```diff
@@ -9,201 +9,1069 @@
...
+                    task: task,
+                    buffer: buffer,
+                    upstreamContinuations: upstreamContinuations,
+                    upstreamsFinished: upstreamsFinished,
+                    downstreamContinuation: nil
+                )
+
+                return .none
+
+            case .upstreamThrew:
```
phausler, line 727: small nit: we rarely use past tense for names of cases. `.upstreamFailure` might be a better name.

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
```diff
@@ -9,201 +9,1069 @@
...
+        /// Actions returned by `upstreamFinished()`.
+        enum UpstreamFinishedAction {
+            /// Indicates that the task and the upstream continuations should be cancelled.
+            case cancelTaskAndUpstreamContinuations(
+                task: Task<Void, Never>,
+                upstreamContinuations: [CheckedContinuation<Void, Error>]
+            )
+            /// Indicates that the downstream continuation should be resumed with `nil` and
+            /// the task and the upstream continuations should be cancelled.
+            case resumeContinuationWithNilAndCancelTaskAndUpstreamContinuations(
```
phausler, line 751: some of these long names and numerous associated types kinda almost infer structness of the associations. Would it be better to change them to structures instead of effectively tuples?

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
```diff
@@ -9,201 +9,1069 @@
...
+                preconditionFailure("Invalid state")
+
+            case .merging(let task, let buffer, let upstreamContinuations, var upstreamsFinished, let .some(downstreamContinuation)):
+                // One of the upstreams finished
+                precondition(buffer.isEmpty, "We are holding a continuation so the buffer must be empty")
+
+                // First we increment our counter of finished upstreams
+                upstreamsFinished += 1
+
+                // We should change the 2 when we get variadic generics
```
phausler, line 772: might be worth it to abstract this to a static inline variable which ought to have the same perf as this. but then give it a name.

phausler (approved):
Overall this looks good, it could use some sprucing up with regards to the locks before merging (id rather not introduce a new locking scheme and keep to the managed critical state stuff if at all possible.

It took me a bit to follow all of the logic but it seems relatively sound (excluding the whole fatal error cases marked in the review).

With changing to the unchecked variants I think we can eek out some more perf!

phausler:
One required item for merging this: it needs to definitely handle rethrowing cases (and we should validate that is true)

Sources/AsyncAlgorithms/Merge2StateMachine.swift
```diff
@@ -0,0 +1,162 @@
...
+        return .second(.success(value), iter)
+      } catch {
+        return .second(.failure(error), iter)
+      }
+    }
+  }
+
+  /// Advances to the next element and returns it or `nil` if no next element exists.
+  mutating func next() async rethrows -> Either? {
+    if Task.isCancelled {
```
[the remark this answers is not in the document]
FranzBusch: This is the old code. I just copied this since other algos were using this and I did not want to change them.

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
```diff
@@ -9,201 +9,1065 @@
...
-  public typealias Element = Base1.Element
-  /// An iterator for `AsyncMerge2Sequence`
-  public struct Iterator: AsyncIteratorProtocol, Sendable {
-    var state: Merge2StateMachine<Base1, Base2>
-    init(_ base1: Base1.AsyncIterator, _ base2: Base2.AsyncIterator) {
-      state = Merge2StateMachine(base1, base2)
+private extension AsyncMerge2Sequence {
+    final class Storage: @unchecked Sendable {
+        /// The lock that protects our state.
+        private let lock = Lock.allocate()
```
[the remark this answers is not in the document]
FranzBusch: I tried to adopt this but `ManagedCriticalState` does only support using a lock with a closure. For the `next` method I would like to manually lock and unlock to be able to hold the lock across the `withUnsafeContinuation` method.

Sources/AsyncAlgorithms/Merge2StateMachine.swift
FranzBusch: Same here, not my code just copied over.

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
FranzBusch: IMO, we should keep these preconditions. The state that I am preconditioning in should never been hit and I haven't been able to force the state machine into these states with my tests. If we ever hit one of those, I consider this a good thing because the assumptions I have made are wrong.

The good thing is this won't become a subtle bug but rather a crash. We have been using this in NIO for a long time and it helped us flush out bugs in all our state machines.

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
FranzBusch: I am using newlines here mostly to make reading it easier not because of screen width. I would prefer to keep that to maintain readability.

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
FranzBusch: Various tests are validating that indirectly. For this precondition to fail we would need to get into a state where we have a suspended call to `next()` and things in the buffer. However, when we call `next` and have things in the buffer we are returning them right away. We are only suspending if the buffer is empty. Furthermore, since we are holding the lock across the `withUnsafeContinuation` call in `next` nothing can get in-between us checking if the buffer is empty and us suspending.

When any of the upstreams produces a new element, it will acquire the lock and then we come into this case. If we have a continuation we are guaranteed to have an empty buffer since we would have never suspended in the first place if we had something in the buffer.

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
FranzBusch: In general I am with you; however, this would introduce just a struct for one case of an action. My personal opinion, is that this is overkill for something internal. If you this is important to you, I am happy to change it though. Your call :)

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
FranzBusch: Agreed, I also wanna change this around since I want to be able to support the merge2 and merge3 algorithm with the same state machine.

FranzBusch:
@phausler I just updated the PR again. To summarise what changed from your initial review:
- My implementation supports rethrowing now
- I split up the implementation into separate files to aid with readability
- I swapped `merge3` to also make use of the new state machine
- I renamed the internal `upstreamThrew` case to `upstreamFailure`
- I added more tests validating various edge cases
- Aligned precondition messages

I replied to some of your feedback inline but I haven't changed the following:
- Use `ManagedCriticalState`: I can't adopt this since I need to be able to manually `lock` and `unlock` in the case when I need to suspend on `next()`
- Remove preconditions: I went through the code again and checked these preconditions and I wasn't able to come up with a test to break them. There might still be an edge case I haven't thought about, but I would rather let this crash then to transition into an unknown state.
- Line breaks: I find them easier to read but if you want me to use a single line I am happy to change (maybe it would be good to get swift format setup in this repo?)

Sources/AsyncAlgorithms/AsyncMerge2Sequence.swift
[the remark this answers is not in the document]
phausler: Holding a lock across a suspension is almost always a really bad idea; it can cause some pretty heinous fall out.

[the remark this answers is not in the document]
FranzBusch: Holding a lock across `with*Continuation` is safe to do since the closure is executed right away. We have been doing this in various places of the NIO ecosystem now.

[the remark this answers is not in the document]
FranzBusch: To further clarify this is not holding a lock across a suspension point but just across the `with*Continuation` call.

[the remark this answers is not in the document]
phausler: so there is only 1 way to do that safely (and it is a very edge case scenario).

```
lock()
await withUnsafeContinuation { continuation in
  self.continuation = continuation
  unlock()
}
```

It is very easy to modify that pattern into some really gnarly failure modes. From a maintainability standpoint id rather make sure there is no other way of doing that.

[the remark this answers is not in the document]
FranzBusch: Right I am using exactly that pattern just with calls to the state inbetween. I agree with you that this needs to be done with care.
However, if don’t hold the lock across that call, the state handling becomes trickier since it introduces a potential interleaving between next and closure of withUnsafeContinuation where everything else can happen

phausler:
@swift-ci please test

FranzBusch:
@phausler Just rebased this PR. Feel free to merge it if you are happy with it!

Issue #192, opened 2022-09-05
[Merge] Improve conception


twittemb:
Hi @phausler, I see an attempt of avoiding creating too much tasks and share Iterators between them in the merge algorithm https://github.com/apple/swift-async-algorithms/pull/185

Although it might work well, the resulting code is pretty complex.

I feel we could leverage async channels as a way of merging elements from several tasks that would consume the upstreams. It would simplify a lot the code. I have a working implementation on a branch that also handles rethrowing. I will open a PR tomorrow to suggest my implementation if you are willing to take a look.

thanks.

FranzBusch:
Thanks for opening this. While I get your concern for simpler code the merge operator is a pretty complex piece.
The main motivation behind my PR is not only to reduce the number of tasks but also to remove the Sendable constraints on the iterator.

At the same time the new implementation is up to 3 times faster since it avoids spawning all these tasks.

Feel free to open a PR but I want to point out some of the requirements you need fulfill;
1. The Iterator must not be Sendable
2. Demand to the upstream must only be issued if the downstream requested an element
3. Cancellation to the upstream must be propagated when the downstream is cancelled
4. Rethrows must be supported

Happy to discuss more here or in the PR.

twittemb:
Hi @FranzBusch 

thanks for those insights. From what I can tell, those requirements are fulfilled by the implementation I have in mind. I would appreciate if you could take a look a the PR I will published tomorrow.

thanks a lot.

twittemb:
Hi @FranzBusch 

thanks for those insights. From what I can tell, those requirements are fulfilled by the implementation I have in mind. I would appreciate if you could take a look a the PR I will published tomorrow.

thanks a lot.

twittemb:
@FranzBusch here is the PR https://github.com/apple/swift-async-algorithms/pull/193
It might need some tweaking and I will ensure the tests all pass repeatedly later today (they all pass individually).

The code is simpler for sure but I decided to reuse and compose AsyncMerge2Sequence in order to make the version with 3 params. I guess we also could do a specific Merge3StateMachine but not sure it worth it.

@phausler what do you think ?

phausler:
[...]

PR #193, opened 2022-09-06
[Merge] optimize tasks creation and remove sendable constraint

twittemb:
Hi

To follow this discussion https://github.com/apple/swift-async-algorithms/issues/192, here is an attempt to reach several goals:

- minimise the task creation (1 per upstream + 1 main for the group)
- have conditional Sendable conformance for AsyncMergeSequences
- make a MergeStateMachine that can be reused with any number of sequences (preparing the field for variadic params)
- be able to use the same MergeStateMachine in `AsyncChunksOfCountOrSignalSequence`
- improve the performances
- reuse the existing locking mechanism

Some hints about the conception:

- I've introduced the concept of `Regulator`. This is a wrapper around an AsyncSequence that can iterates over it and suspend before calling next until it is resumed by an external entity. I think this is something we can leverage also in `AsyncCombineLatestSequence` where we will also want to avoid creating too much tasks. The Regulator has its own state, not shared with any one else, there is no synchronisation point with the outside.
- The MergeStateMachine is no more specialised with several `AsyncSequence` generic types but only by the `Element` type, which opens the door to variadic params. The MergeStateMachine has its own state also, independent from the upstreams.

Performance wise, this is what I have measured on my M1 Ultra:

Currently:
- Merge2: 150 000 events/s
- Merge3: 115 000 events/s

My PR:
- Merge2: 320 000 events/s
- Merge3: 260 000 events/s

I've added a Merge4 operation in the throughput test just to show that it is now possible to merge any number of AsyncSequence.

There is a significant improvement. @FranzBusch might have even better results with his implementation and perhaps we can improve mine.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,9 +16,14 @@ let package = Package(
     .library(name: "_CAsyncSequenceValidationSupport", type: .static, targets: ["AsyncSequenceValidation"]),
     .library(name: "AsyncAlgorithms_XCTest", targets: ["AsyncAlgorithms_XCTest"]),
   ],
-  dependencies: [],
+  dependencies: [
+    .package(url: "https://github.com/apple/swift-collections.git", .upToNextMajor(from: "1.0.3"))
+  ],
   targets: [
-    .target(name: "AsyncAlgorithms"),
+    .target(
+      name: "AsyncAlgorithms",
+      dependencies: [.product(name: "Collections", package: "swift-collections")]
+    ),
     .target(
       name: "AsyncSequenceValidation",
       dependencies: ["_CAsyncSequenceValidationSupport", "AsyncAlgorithms"]),
--- a/Sources/AsyncAlgorithms/AsyncChannel.swift
+++ b/Sources/AsyncAlgorithms/AsyncChannel.swift
@@ -28,7 +28,7 @@ public final class AsyncChannel<Element: Sendable>: AsyncSequence, Sendable {
     init(_ channel: AsyncChannel<Element>) {
       self.channel = channel
     }
-    
+
     /// Await the next sent element or finish.
     public mutating func next() async -> Element? {
       guard active else {
@@ -116,7 +116,7 @@ public final class AsyncChannel<Element: Sendable>: AsyncSequence, Sendable {
   
   /// Create a new `AsyncChannel` given an element type.
   public init(element elementType: Element.Type = Element.self) { }
-  
+
   func establish() -> Int {
     state.withCriticalRegion { state in
       defer { state.generation &+= 1 }
@@ -152,7 +152,7 @@ public final class AsyncChannel<Element: Sendable>: AsyncSequence, Sendable {
           }
           return UnsafeResumption(continuation: send, success: continuation)
         case .awaiting(var nexts):
-          if nexts.update(with: Awaiting(generation: generation, continuation: continuation)) != nil {
+            if nexts.update(with: Awaiting(generation: generation, continuation: continuation)) != nil {
             nexts.remove(Awaiting(placeholder: generation))
             cancelled = true
           }
--- a/Sources/AsyncAlgorithms/AsyncChunksOfCountOrSignalSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncChunksOfCountOrSignalSequence.swift
@@ -60,34 +60,50 @@ public struct AsyncChunksOfCountOrSignalSequence<Base: AsyncSequence, Collected:
 
   public typealias Element = Collected
 
+  enum Either {
+    case first(Base.Element)
+    case second(Signal.Element)
+  }
+
   /// The iterator for a `AsyncChunksOfCountOrSignalSequence` instance.
   public struct Iterator: AsyncIteratorProtocol, Sendable {
     let count: Int?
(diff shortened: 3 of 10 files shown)


FranzBusch:
Hey @twittemb,

I just had a brief look over your last push and your are getting closer and closer to what I have done in the other PR. However I spotted a couple of things that are different and won't work like stuffing the bases into an Array. In the end, we already have what we want in the other PR and it just requires the last final touches.

I would love to understand what you want to achieve that the other PR doesn't achieve?

twittemb:
> Hey @twittemb,
> 
> I just had a brief look over your last push and your are getting closer and closer to what I have done in the other PR. However I spotted a couple of things that are different and won't work like stuffing the bases into an Array. In the end, we already have what we want in the other PR and it just requires the last final touches.
> 
> I would love to understand what you want to achieve that the other PR doesn't achieve?

Hi @FranzBusch 

I've updated the description to explain my intentions. My intent was to have some components that are reusable (like I said, AsyncCombineLatestSequence will have to be refactored as well) and performant and that are ready for variadic params without having to duplicate the code every time.

Could you explain the "array" thing please ? (by the way the implementation of the variadic params here is just a "demo" to show the intent)

FranzBusch:
> > Hey @twittemb,
> > I just had a brief look over your last push and your are getting closer and closer to what I have done in the other PR. However I spotted a couple of things that are different and won't work like stuffing the bases into an Array. In the end, we already have what we want in the other PR and it just requires the last final touches.
> > I would love to understand what you want to achieve that the other PR doesn't achieve?
> 
> Hi @FranzBusch
> 
> I've updated the description to explain my intentions. My intent was to have some components that are reusable (like I said, AsyncCombineLatestSequence will have to be refactored as well IMO) and performant and that are ready for variadic params without having to duplicate the code every time.
> 
> Could you explain the "array" thing please ? (by the way the implementation of the variadic params here is just a "demo" to show the intent)

While I fully understand that you wanna abstract things away (this is something we programmers always want to do quite quickly) I personally don't believe this is the right approach for this package with the exception of some algorithms that can be trivially implemented on top of others.

The reason why I think this is a no-goal is that abstractions often incur some kind of cost. Either performance, size or usability. Furthermore, coming up with arbitrary abstractions makes it harder to understand what the implementation does and reason if it does the right thing. In the end, one should be able to look at the implementation of the code and go trough the various states that an algorithm can have and understand how it reacts. In your implementation, I find it personally very hard to follow what goes on since there are types in the mix that are not very self-explaining and there is no comments along the way.

On the last note why using an Array is bad. The array results in a heap allocation which makes this algorithms worse performance wise. My implementation is also ready for variadic generics we just need to replace the abstract state machine's generic types with a variadic generic.

twittemb:
Hi @FranzBusch 

Thanks for you detailed answer. When I introduced the `Regulator`, my intent was not the randomly abstract things but to provide a tool that could be reused in several situations (trying to not fall in the pit of early optimisation). I can imagine that a sequence that can be suspended/resumed at will can be a useful tool (merge and combineLatest are 2 good examples IMO). In the test package we also have a `GatedSequence` as a test tool and perhaps we could merged those implementations.

As an OpenSource package I think it is also important for contributors to suggest new things based on the tools we could provide (the existing locking mechanism is a nice example), but that's just my opinion and might not be relevant (also it is my first time contributing to an Apple repo) :-)

I will cancel this PR and the associated Issue to remove the noise. My last concern was about `AsyncChunksOfCountOrSignalSequence` and the fact that it still uses the old Merge2StateMachine from what I understand. Is this something you plan to address ? I feel it would be bad to maintain 2 different implementations, and the old one is not as performant as yours.

@phausler one last though about this ?

FranzBusch:
> Hi @FranzBusch
> 
> Thanks for you detailed answer. When I introduced the `Regulator`, my intent was not the randomly abstract things but to provide a tool that could be reused in several situations (trying to not fall in the pit of early optimisation). I can imagine that a sequence that can be suspended/resumed at will can be a useful tool (merge and combineLatest are 2 good examples IMO). In the test package we also have a `GatedSequence` as a test tool and perhaps we could merged those implementations.
> 
> As an OpenSource package I think it is also important for contributors to suggest new things based on the tools we could provide (the existing locking mechanism is a nice example), but that's just my opinion and might not be relevant (also it is my first time contributing to an Apple repo) :-)

Don't get me wrong, we are super happy with any contributions we get and you already have provided some very nice improvements throughout this package. Please keep them coming, we really appreciate them!

> I will cancel this PR and the associated Issue to remove the noise. My last concern was about `AsyncChunksOfCountOrSignalSequence` and the fact that it still uses the old Merge2StateMachine from what I understand. Is this something you plan to address ? I feel it would be bad to maintain 2 different implementations, and the old one is not as performant as yours.

You are right there is still more to migrate but I wanted to keep the PR focused and `chunked` hasn't been pitched yet. Though as you said we really need to reimplement this!

There are also more algorithms in here that currently constrain the `Sendable` requirement of the `AsyncIterator` which we should revisit and understand if they are really required.

twittemb:
Thanks for your kind words. Always happy to help, and will continue to try to. looking forward to you next implementations.
Have a nice one.

PR #194, opened 2022-09-07
Convert the guides for debounce and throttle into a proposal

phausler:
This proposal formalizes rate limiting algorithms for debouncing and throttling.

Read the full proposal [here](https://github.com/phausler/swift-async-algorithms/blob/1b7c4e9e63daa27aa7566c61980723d5b279031b/Evolution/NNNN-rate-limits.md)

### DIFF
--- a/Evolution/NNNN-rate-limits.md
+++ b/Evolution/NNNN-rate-limits.md
@@ -1,0 +1,159 @@
+# Rate Limiting
+
+* Proposal: [SAA-NNNN]()
+* Authors: [Philippe Hausler](https://github.com/phausler)
+* Status: **Implemented**
+* Implementation: 
+[
+[Source](https://github.com/apple/swift-async-algorithms/blob/main/Sources/AsyncAlgorithms/AsyncDebounceSequence.swift) |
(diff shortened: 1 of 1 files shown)


FranzBusch:
@phausler Debounce ought to implementable without requiring the iterator to be Sendable. Have we looked into this?

phausler:
@FranzBusch the Sendable requirement definitely complicates the code considerably (again I think falls into the "this package should do it so other folks shouldn't" category). I am working on a version that reduces the task creation overhead and the sendable requirement to iterators.... but it makes the merge PR look simple in comparison... Sendable should be something part of the review. 

@twittemb Good call, I think half of that report is resolved now due to fixes in clock, I have not yet had a chance to verify other parts of it. The behavior should not be a full predication on reviewing this however.

FranzBusch:
> @FranzBusch the Sendable requirement definitely complicates the code considerably (again I think falls into the "this package should do it so other folks shouldn't" category). I am working on a version that reduces the task creation overhead and the sendable requirement to iterators.... but it makes the merge PR look simple in comparison... Sendable should be something part of the review.

I am with you that it complicates things but we really should fix this since it has significant usability downsides. Like you said "let's get it right once in this package". Happy to help with this!

phausler:
@swift-ci please test

PR #198, opened 2022-09-11
Remove `AsyncIterator: Sendable` requirement from debounce

FranzBusch:
# Motivation
The current implementation of `AsyncDebounceSequence` requires the base `AsyncIterator` to be `Sendable`. This is causing two problems:

1. It only allows users to use `debounce` if their `AsyncSequence.AsyncIterator` is `Sendable`
2. In `debounce` we are creating a lot of new `Task`s and reating `Task`s is not cheap.

My main goal of this PR was to remove the `Sendable` constraint from `debounce`.

# Modification
This PR overhauls the implementation of `debounce` and aligns it with the implementation of the open `merge` PR https://github.com/apple/swift-async-algorithms/pull/185 . The most important changes are this:
- I removed the `Sendable` requirement from the base sequences `AsyncIterator`.
- Instead of creating new Tasks for the sleep and for the upstream consumption. I am now creating one Task and manipulate it by signalling continuations
- I am not cancelling the sleep. Instead I am recalculating the time left to sleep when a sleep finishes.

# Result
In the end, this PR swaps the implementation of `AsyncDebounceSequence` and drops the `Sendable` constraint and passes all tests. Furthermore, on my local performance testing I saw up to 150% speed increase in throughput.


FranzBusch:
This passes the tests but I still need to do some cleanup and check all comments. Furthermore, I wanna add more tests for the edge-cases

phausler:
So far this looks like a nice improvement. I will take a fine tooth comb to it on Monday.

Sources/AsyncAlgorithms/Debounce/AsyncDebounceSequence.swift
```diff
@@ -0,0 +1,115 @@
...
+    public func debounce(for interval: Duration, tolerance: Duration? = nil) -> AsyncDebounceSequence<Self, ContinuousClock> where Self: Sendable {
+        self.debounce(for: interval, tolerance: tolerance, clock: .continuous)
+    }
+}
+
+/// An `AsyncSequence` that emits the latest element after a given quiescence period
+/// has elapsed.
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct AsyncDebounceSequence<Base: AsyncSequence, C: Clock>: Sendable where Base: Sendable {
+    /// This class is needed to hook the deinit to observe once all references to the ``AsyncDebounceSequence`` are dropped.
```
phausler, line 32: why is the lifespan of the AsyncSequence (the plan on HOW to do things) going to terminate the iteration? im not sure I follow on why this is appropriate behavior.

Sources/AsyncAlgorithms/Debounce/DebounceStateMachine.swift
```diff
@@ -0,0 +1,757 @@
...
+    enum IteratorDeinitializedAction {
+        /// Indicates that the `Task` needs to be cancelled and
+        /// the upstream and clock continuation need to be resumed with a `CancellationError`.
+        case cancelTaskAndUpstreamAndClockContinuations(
+            task: Task<Void, Never>,
+            upstreamContinuation: UnsafeContinuation<Void, Error>?,
+            clockContinuation: UnsafeContinuation<C.Instant, Error>?
+        )
+        /// Indicates that nothing should be done.
+        case none
```
phausler, line 114: almost seems like it is an optional struct and not an enum ;)

Not a big deal but is sus.

Sources/AsyncAlgorithms/Debounce/DebounceStorage.swift
```diff
@@ -0,0 +1,311 @@
...
+        }
+    }
+
+    func next() async rethrows -> Element? {
+        // We need to handle cancellation here because we are creating a continuation
+        // and because we need to cancel the `Task` we created to consume the upstream
+        return try await withTaskCancellationHandler {
+            // We always suspend since we can never return an element right away
+
+            self.lock.lock()
```
phausler, line 265: this is the only straddling lock, I wonder if we can wrap this up into something within the managed critical state. so that for future use it is ONLY the safe edge case being used here.

Sources/AsyncAlgorithms/Debounce/DebounceStorage.swift
```diff
@@ -0,0 +1,311 @@
...
+            return try await withUnsafeThrowingContinuation { continuation in
```
phausler, line 266: im VERY surprised this actually works as rethrowing. I would have expected that you need to use a `Result` to `_rethrowGet()` the value.

twittemb: I think this is related to the issue I've spotter here -> https://github.com/apple/swift/issues/60821.
We can bypass the mechanism by wrapping the call inside any rethrowing function (like `withTaskCancellationHandler`)

phausler: yea that really seems like a compiler soundness hole to me. If at all possible we should avoid relying on it and paint everything into one common handler; it might be worth a secondary investigation of which approach; replicating this bug into a method to with a closure or using the rethrow-get both simulate a un-yet-implemented `rethrow(unchecked)` compiler feature.

Sources/AsyncAlgorithms/Debounce/AsyncDebounceSequence.swift
FranzBusch: You are right that doesn't make sense here. This is some of the behaviour I inherited from the implementation of Source `AsyncSequence`s. I changed it there since `debounce` should inherit the _castiness_ from it's base `AsyncSequence`

Sources/AsyncAlgorithms/Debounce/DebounceStorage.swift
FranzBusch: Yeah I found that soundness hole as well and wanted to check that actually. I just changed this to use `Result<Element?, Error>` instead.

phausler:
@swift-ci please test

phausler (approved):
Overall this looks fantastic, it seems that this is definitely both a perf and correctness boon!

Sources/AsyncAlgorithms/Locking.swift
```diff
@@ -87,6 +87,27 @@ internal struct Lock {
...
+
+    /// Acquire the lock for the duration of the given block.
+    ///
+    /// This convenience method should be preferred to `lock` and `unlock` in
+    /// most situations, as it ensures that the lock will be released regardless
+    /// of how `body` exits.
+    ///
+    /// - Parameter body: The block to execute while holding the lock.
+    /// - Returns: The value returned by the block.
+    func withLock<T>(_ body: () throws -> T) rethrows -> T {
```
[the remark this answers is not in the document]
FranzBusch: Yep, removed them!

Sources/AsyncAlgorithms/Debounce/DebounceStorage.swift
```diff
@@ -0,0 +1,296 @@
...
+                        self.startTask(
+                            stateMachine: &$0,
+                            base: base,
+                            downstreamContinuation: continuation
+                        )
+
+                    case .resumeUpstreamContinuation(let upstreamContinuation):
+                        // This is signalling the upstream task that is consuming the upstream
+                        // sequence to signal demand.
+                        upstreamContinuation?.resume(returning: ())
```
phausler, line 72: hmm these continuations are being resumed inside the critical region; doesn't that mean that we could potentially get a recursion in the lock?

Sources/AsyncAlgorithms/Debounce/DebounceStorage.swift
```diff
@@ -0,0 +1,296 @@
...
+
+                    case .resumeUpstreamAndClockContinuation(let upstreamContinuation, let clockContinuation, let deadline):
+                        // This is signalling the upstream task that is consuming the upstream
+                        // sequence to signal demand and start the clock task.
+                        upstreamContinuation?.resume(returning: ())
```
phausler, line 77: I think perhaps we need to make sure to return these back out and resume them outside of the lock.

phausler:
I think the task creation feels right this way; basically I think the rule is that if work is to be done it isn't done until thee first call to next - signifying demand.

`myAsyncSeq.makeAsyncIterator()` really should be pretty low in work.

Sources/AsyncAlgorithms/Debounce/DebounceStorage.swift
FranzBusch: Resuming continuations while hold the lock is safe. From the docs of `resume`
> After resume enqueues the task, control immediately returns to the caller. The task continues executing when its executor is able to reschedule it.

This plays out to the resumed job to be enqueued in the appropriate executor and we will continue right away with our code and drop the lock.

Sources/AsyncAlgorithms/Debounce/DebounceStorage.swift
```diff
@@ -0,0 +1,296 @@
...
+        return try await withTaskCancellationHandler {
+            // We always suspend since we can never return an element right away
+
+            let result: Result<Element?, Error> = await withUnsafeContinuation { continuation in
+                self.stateMachine.withCriticalRegion {
+                    let action = $0.next(for: continuation)
+
+                    switch action {
+                    case .startTask(let base):
+                        self.startTask(
```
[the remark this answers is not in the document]
FranzBusch: This is also safe to do since we are starting the `Task` while holding the lock. If we are in a single threaded env the new `Task` will get created but it will never make any progress until the current job (which is calling `next()`) is finished.

If we are not holding the lock across the start of the `Task` we get weird race conditions where the upstream consumption child tasks could call out to the state machine before we even transitioned from `initial` which I want to avoid and would cause unnecessary complexity in the state handling.

Sources/AsyncAlgorithms/Debounce/DebounceStorage.swift
FranzBusch: See above comment but this ought to be totally safe

Sources/AsyncAlgorithms/Debounce/DebounceStorage.swift
phausler: hmm that means perhaps we have some work to do for some optimizations for the entire framework because I was under I guess the mistaken presumption that there was an edge case where it could be resumed such that it becomes recursive

PR #208, opened 2022-10-10
AsyncSubject/AsyncThrowingSubject

rustle:
Starting a Subject implementation exploration for moving values from a non-async context into an async context in sequence.


twittemb:
Hi @rustle 

I allow my self to post a comment here since I’ve made the same kind of implementation in a pitch on the forum -> https://forums.swift.org/t/pitch-async-buffered-channel/59854

The outcome seemed to be that it was too close to `AsyncStream` and that having a factory function like https://github.com/sideeffect-io/AsyncExtensions/blob/main/Sources/Creators/AsyncStream%2BPipe.swift should be enough (and could be pitched to the standard library).

Keithbreadley (approved):

PR #209, opened 2022-10-11
Simplify chunks algorithm by re-using merge and composition

phausler:
This approach uses a fancy pants trick by composing merge with map and chain. This accounts for the trailing termination by appending to the base sequence a tri-state on the "either" such that it incurs a termination. Effectively it indicates the chunks algorithm is a reductions composed with merge and chain.


Sources/AsyncAlgorithms/AsyncChunksOfCountOrSignalSequence.swift
```diff
@@ -56,35 +56,54 @@ extension AsyncSequence {
...
-          case .second(_):
-            if result != nil {
-              return result
-            }
+        case .element(let element):
+          if result == nil {
+            result = Collected()
+          }
+          result!.append(element)
+          if result?.count == count {
```
kperryua, line 97: We can force-unwrap `result!` instead of optional chain this here. (This was extant, I see).

kperryua (commented):
This looks good from a correctness standpoint. However, I'm interested in seeing the throughput of this implementation compared to the prior implementation that has a reimplementation of `Merge2StateMachine` that takes advantage of the single-Task changes to `merge`. I recall the use of `Merge2StateMachine` resulted in much more effective inlining than calling into `merge.

PR #210, opened 2022-10-11
Relax Sendable conditional conformance for AsyncJoinedBySeparatorSequence

phausler:
The requirement was overly strict for AsyncJoinedBySeparatorSequence and can be safely relaxed.


FranzBusch (approved):
Nice!

PR #211, opened 2022-10-12
Simplify the implementation of `zip`

FranzBusch:
# Motivation
We recently merged the PR that corrected the `Sendable` constraint of the. After thinking about this some more, I realised we overly complicated the implementation here. The great thing about `zip` is that it is actually trivial and we don't have to race any of the upstream sequences against each other. As soon as I realised that, I knew we can completely get rid of the `Task` and just implement this on top of a task group.

# Modification
Remove the complicated state machine implementation and replace it with a straight forward task group.

# Result
This makes the implementation of `zip` tremendously shorter. Furthermore, it improves throughout by roughly 1.5 times.


FranzBusch:
cc @twittemb I think that makes a lot of our stuff easier here. An important lesson learned for me: Take a step back and look at the algo again, if there is nothing time based or racing of upstreams involved, it might be implemented on the current Concurrency primitives.

FranzBusch:
> I think we can reuse my zip state machine implementation and adapt it easily to `combineLatest` as it is a form of zipping. Currently the implementation of combineLatest uses Tasks + a constraint on the iterator to be Sendable. If you give me a little bit of time I think I can come up with a replacement. What do you think ?

I have this already in progress and will open a PR later today or tomorrow for this.

FranzBusch:
Actually, this isn't working after all. I skipped one important setting here which is enabling `--warn-concurrency` and obviously we are sending the iterator across child tasks here. Which doesn't work. Closing this now.

twittemb:
> Actually, this isn't working after all. I skipped one important setting here which is enabling `--warn-concurrency` and obviously we are sending the iterator across child tasks here. Which doesn't work. Closing this now.

I’m implementing the changes you originally requested in my `zip` PR to prepare for variadic params + improve the naming (getting rid of the zip2 state machine). I should submit something in the coming days (probably next week).

FranzBusch:
> I’m implementing the changes you originally requested in my `zip` PR to prepare for variadic params + improve the naming (getting rid of the zip2 state machine). I should submit something in the coming days (probably next week).

Please don't spend time on this. I have done this already for my work on `combineLatest`. That's why we merged the PR already. Sorry, for taking this over but we want to straighten out the last `Sendable` things in the next few days.

twittemb:
> > I’m implementing the changes you originally requested in my `zip` PR to prepare for variadic params + improve the naming (getting rid of the zip2 state machine). I should submit something in the coming days (probably next week).
> 
> Please don't spend time on this. I have done this already for my work on `combineLatest`. That's why we merged the PR already. Sorry, for taking this over but we want to straighten out the last `Sendable` things in the next few days.

Ok sure. No problem 😉  the important thing is that it will be done !

PR #212, opened 2022-10-12
Remove `Sendable` constraint for `AsyncAdjacentPairsSequence`

FranzBusch:
# Motivation
We want to remove any `Sendable` constraints on iterators since they ought to not be passed between `Task`s.

# Modification
Explicitly mark the `AsyncAdjacentPairsSequence` to be not `Sendable`.

# Result
Nobody can make this `Sendable`.

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncAdjacentPairsSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncAdjacentPairsSequence.swift
@@ -83,5 +83,7 @@ extension AsyncSequence {
   }
 }
 
-extension AsyncAdjacentPairsSequence: Sendable where Base: Sendable, Base.Element: Sendable, Base.AsyncIterator: Sendable { }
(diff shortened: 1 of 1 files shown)


phausler:
Assume `Base.Iterator: Sendable`

`Sendable` requires all usages across tasks to be well formed

usage of an iterator across tasks can result in two tasks deriving the same value (broadcast) or deriving subsequent values (multicast)

any conditional conformance of `Sendable` must preserve the usage form (broadcast or multicast) characteristic

given the same values upstream the adjacent pairs downstream by definition should be the same
given an upstream of `1, 2, 3, 4` results in `(1, 2), (2, 3), (3, 4)`

if the upstream is broadcast then the adjacent pairs would be the same

if the upstream is multicast and provides `1, 2, 3, 4` it can be split into two via one producing `1, 3` and another producing `3, 4` then the downstream would be `(1, 3)` and `(3, 4)`

does that latter violate the multicast characteristic?

phausler:
I was able to prototype a version of a broadcast algorithm that does not require the iterators to be sendable - so this change is fine!

phausler (approved):

PR #213, opened 2022-10-12
Remove `Sendable` constraint for `AsyncChain2Sequence`

FranzBusch:
# Motivation
We want to remove any `Sendable` constraints on iterators since they ought to not be passed between `Task`s.

# Modification
Explicitly mark the `AsyncChain2Sequence` to be not `Sendable`.

# Result
Nobody can make this `Sendable`.


phausler:
I think this conditional conformance actually works in all cases; multicast will still maintain that behavior, broadcast will also maintain that behavior too. All that is stored per state is the iterators and derived state from those.

phausler (approved):
Same conclusion as #212

Issue #110, opened 2022-03-26
Multi-consumption of an `AsyncSequence`


ABridoux:
Hi, I have been reading the documentation for this repository and it left me wondering: is this planned to introduce a feature to allow several tasks to consume the same `AsyncSequence`?

_If this is not the right place to offer ideas, please let me know. I don't think this would fit in a proposal._

## Context
I recently had to stream a file using `URLSession` and two pieces of code were interested in the streamed values. Rather than starting a stream twice which would not be efficient, I wanted to allow the stream to be consumed by several tasks. The data that were already streamed would be sent when a new consumption is set up.

## Basic implementation
I tried to implement such a solution (in a [gist](https://gist.github.com/ABridoux/913dfb95fabc22048b0a1c566059105a)) that will use a reduce function on the already emitted values and emit the result when a new task starts consuming the sequence. It's far from being perfect but I think it might help understanding the idea.

<details><summary>Implementation</summary>
<p>

```swift
struct ReducedReplayAsyncStream<Element> {

    typealias Reduce = (_ partialResult: inout Element, _ nextResult: Element) -> Void

    private let storage: _Storage
    private var originalStream: AsyncStream<Element>

    init(
        bufferingPolicy limit: AsyncStream<Element>.Continuation.BufferingPolicy = .unbounded,
        initialResult: Element,
        reduce: @escaping Reduce,
        build: (AsyncStream<Element>.Continuation) -> Void
    ) {
        originalStream = AsyncStream(Element.self, bufferingPolicy: limit, build)
        storage = _Storage(stored: initialResult, reduce: reduce)
    }

    private func makeStream() -> AsyncStream<Element> {
        AsyncStream<Element> { continuation in
            Task {
                var isFirst = false
                if await !storage.didStart {
                    await storage.setDidStart(true)
                    isFirst = true
                    startConsumingOriginalStream()
                }

                if !isFirst {
                    await continuation.yield(storage.stored)
                }
                await storage.appendContinuation(continuation)
            }
        }
    }

    private func startConsumingOriginalStream () {
        Task {
            for await value in originalStream {
                await storage.updateWith(value: value)
            }
            await storage.continuations.forEach { $0.finish() }
        }
    }
}

extension ReducedReplayAsyncStream {

    private actor _Storage {
        private let reduce: ReducedReplayAsyncStream.Reduce

        var didStart = false
        var stored: Element
        var continuations: [AsyncStream<Element>.Continuation] = []

[...]

PR #214, opened 2022-10-14
[WIP] Broadcast algorithm

phausler:
This algorithm allows for the broadcasting of values to multiple consumers from one source. 

Multiple instances from the same broadcast will share the values produced from iteration. These also share failures that are thrown or non-throwing termination. Cancellation is handled on a per instance basis, but if all iterations are cancelled and there are no more references to the base iteration then the base iteration is cancelled. 

There are bits of room for improvement: currently each side contains a buffer to the elements yet to be consumed. This could be changed to an index into a shared buffer, however the algorithm to share that is quite complex and could be a bit tricky to implement. So as an initial draft this is being left as `O(N * M)` storage where N is the number of active iterations and `M` is the number of elements that have yet to be consumed. In degenerate cases this can be looked at as `O(N^2)` storage, however in ideal cases it is `O(1)`. The shared buffer can bring it down to degenerate cases being `O(N)` and the normal case being `O(1)`.

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncBroadcastSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncBroadcastSequence.swift
@@ -1,0 +1,246 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+import DequeModule
+
+extension AsyncSequence where Self: Sendable, Element: Sendable {
+  public func broadcast() -> AsyncBroadcastSequence<Self> {
+    AsyncBroadcastSequence(self)
+  }
+}
+
+public struct AsyncBroadcastSequence<Base: AsyncSequence>: Sendable where Base: Sendable, Base.Element: Sendable {
+  struct State : Sendable {
+    enum Terminal {
+      case failure(Error)
+      case finished
+    }
+    
+    struct Side {
+      var buffer = Deque<Element>()
+      var terminal: Terminal?
+      var continuation: UnsafeContinuation<Result<Element?, Error>, Never>?
+      
+      mutating func drain() {
+        if !buffer.isEmpty, let continuation {
+          let element = buffer.removeFirst()
+          continuation.resume(returning: .success(element))
+          self.continuation = nil
+        } else if let terminal, let continuation {
+          switch terminal {
+          case .failure(let error):
+            self.terminal = .finished
+            continuation.resume(returning: .failure(error))
+          case .finished:
+            continuation.resume(returning: .success(nil))
+          }
+          self.continuation = nil
+        }
+      }
+      
+      mutating func cancel() {
+        buffer.removeAll()
+        terminal = .finished
+        drain()
+      }
+      
+      mutating func next(_ continuation: UnsafeContinuation<Result<Element?, Error>, Never>) {
+        assert(self.continuation == nil) // presume that the sides are NOT sendable iterators...
+        self.continuation = continuation
+        drain()
+      }
+      
+      mutating func emit(_ result: Result<Element?, Error>) {
+        switch result {
+        case .success(let element):
+          if let element {
+            buffer.append(element)
+          } else {
+            terminal = .finished
+          }
+        case .failure(let error):
+          terminal = .failure(error)
+        }
+        drain()
+      }
+    }
+    
+    var id = 0
+    var sides = [Int: Side]()
+    
+    init() { }
+    
+    mutating func establish() -> Int {
+      defer { id += 1 }
+      sides[id] = Side()
+      return id
+    }
+    
+    static func establish(_ state: ManagedCriticalState<State>) -> Int {
+      state.withCriticalRegion { $0.establish() }
+    }
+    
+    mutating func cancel(_ id: Int) {
+      if var side = sides.removeValue(forKey: id) {
+        side.cancel()
+      }
+    }
+    
+    static func cancel(_ state: ManagedCriticalState<State>, id: Int) {
+      state.withCriticalRegion { $0.cancel(id) }
+    }
+    
+    mutating func next(_ id: Int, continuation: UnsafeContinuation<Result<Element?, Error>, Never>) {
+      sides[id]?.next(continuation)
+    }
+    
+    static func next(_ state: ManagedCriticalState<State>, id: Int) async -> Result<Element?, Error> {
+      await withUnsafeContinuation { continuation in
+        state.withCriticalRegion { $0.next(id, continuation: continuation) }
+      }
+    }
+    
+    mutating func emit(_ result: Result<Element?, Error>) {
+      for id in sides.keys {
+        sides[id]?.emit(result)
+      }
+    }
+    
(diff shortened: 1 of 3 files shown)


twittemb:
Hi @phausler 

Happy we are beginning to introduce this kind of operator :-).

I have a few question regarding this draft though:
- it looks like an equivalent to the Combine `share` operator (multicast + auto start), meaning that future consumers will miss the initial values, is that the intent ?
- shouldn’t we think about a family of “sharing” operators with some options (like replaying values). Some could use buffers, some could rely on `AsyncChannel` to enforce all consumers to consume before requesting the next elements ?
- From what I understand, the base is iterated over in a dedicated task no matter the demand from the consumers. Am I right ? Isn’t it weird ?

Thanks.

phausler:
So it is the intent that prior values are NOT buffered for later consumption when new downstreams are attached (else the best case storage would be O(N^2)... which imho is unexpected: and if you need that then you can slap a buffer on it maybe?)

This is by no means the only one of this category: I'm sure there are either composable variants or other distinct algorithms in the same "multicast" family. This is just the one I see more commonly requested/used.

I think (with some extra complexity) I can modify this to respond to per element demand to pump the iterator along; that is likely tied to the consolidation of the buffers I was mentioning in the pr description. This is still a WIP so my next steps are to make that incremental demand and shared buffers. (And measure the impact/complexity of storage)

FranzBusch:
Nice start! I agree that we should get rid of the `Task` in here and it ought to be possible. This is definitely one of the more interesting algorithms implementation wise :D

Sources/AsyncAlgorithms/AsyncBroadcastSequence.swift
```diff
@@ -0,0 +1,246 @@
...
+        do {
+          for try await element in base {
+            State.emit(state, result: .success(element))
+          }
+          State.emit(state, result: .success(nil))
+        } catch {
+          State.emit(state, result: .failure(error))
+        }
+      }
+    }
```
FranzBusch, line 146: One thing I just noticed after the discussion in the forums around deferred. This is actually creating multiple upstream iterators if I understand this code correctly. IMO we really need to avoid this otherwise this algorithm is not capable of transforming a unicast `AsyncSequence` into a broadcasted one. @phausler please correct if I am misunderstanding the code here

phausler: no, this is creating 1 singular task to iterate so only 1 upstream iterator is made.

phausler:
@tcldr this might be of interest, a slightly different approach.

tcldr:
@phausler Yes, this served as the inspiration for the work on `shared`! I originally intended to see if I could adapt this by adding history and a disposing iterator and then fell down the rabbit hole. :) (I borrowed your original test cases, too.)

The thing that begun to interest me is how this would be used with something like `AsyncChannel`. As this algorithm uses internal buffers, the utility of AsyncChannel is lost. I realised that if back pressure was to be transmitted from the site of consumption to the site of production, back pressure would need to be maintained _throughout_ the pipeline.

The other thing I realised is that while a back pressure supporting algorithm can be made to work in a ‘depressurised’ way (by chaining a buffer), the inverse _isn’t_ possible.

So I thought, maybe if I create a back pressure supporting multicast algorithm, it could be used as a primitive to support _both_ use-cases: ones where back-pressure is a requirement and as well as those use cases where it isn’t (by chaining buffers).

Sources/AsyncAlgorithms/AsyncBroadcastSequence.swift
```diff
@@ -0,0 +1,246 @@
...
+        self.state = state
+        self.iteration = iteration
+        self.id = State.establish(state)
+      }
+      
+      deinit {
+        State.cancel(state, id: id)
+        if iteration.isKnownUniquelyReferenced() {
+          Iteration.cancel(iteration)
+        }
```
kabiroberai, line 208: I might be missing something here, but if the last two iterators enter `deinit` at the same time (and both perform the `isKnownUniquelyReferenced` check before releasing `iteration`), could that trigger a race condition such that neither ends up calling `Iteration.cancel`?

Sajjon:
I would love to use this soon, do you plan to continue working on this in January @phausler ? :)

Keithbreadley (approved):

Sajjon:
How does this PR relate to https://github.com/apple/swift-async-algorithms/pull/242 ?

Sajjon:
@FranzBusch almost 18 months later, any update? :) I think multicast is a crucial missing piece in Async Swift.

dehrom:
@FranzBusch it's really useful feature 🙏

BrentMifsud:
any movement on this?

Kyome22:
In macOS, it is common and frequent to want to use one data source in multiple windows.
Therefore, I am eagerly awaiting the implementation of a broadcast algorithm like the one proposed here.
It is required for replacing Combine's CurrentValueSubject with Swift Concurrency.

Sources/AsyncAlgorithms/AsyncBroadcastSequence.swift
jdberry: I'm clearly being dense, because it also looks to me like this would create a new Task, at least each time `broadcast` is called. And for each `Task` there would be a separate iterator of the base sequence.

yanniks:
@FranzBusch any update here? We also think that this is a crucial feature. Or is the whole `swift-async-algorithms` project abandoned and no longer maintained? 😞

parkera:
There are two main reasons why this particular PR hasn't been merged. First, it is an API with a lot of potential for confusion (this was also the case with the CurrentValueSubject in Combine). It deserves a detailed discussion and exploration. Second, we think that many of the recent advances in the language around `~Copyable` and `~Escapable` have a lot of potential to make API in this space much easier to understand. That is starting to coalesce in the next few Swift releases, but there are still many open questions around sequences and escapability that need resolution in the standard library.

For those reasons, it makes sense for this project to remain fairly conservative on adding major new APIs until we can be sure that it will fit in with the ecosystem as a whole.

jshier:
@parkera Given this library is semantically versioned and shipped outside platform toolchains, such a position seems unnecessarily conservative. In the first case the library can simply ship a `typealias` if a better name comes up (which only requires a minor release). As to the second, it sounds like a major refactor anyway, which should require a major release. I think the ecosystem would benefit much more from getting these APIs early, even if there's a bit of churn later, rather than waiting **years** for it to appear at all.

Sajjon:
I fully agree with @jshier 

Given the high demand on this particular feature - it is an important piece of functionality which community craves - isn't it better to vendor something good now which can be perfected later? 

Furthermore, this particular feature (and many of the already shipped features of this lib) is hard to get right... so either we risk the entire ecosystem introducing bugs because of incorrect implementation of this feature because Apple is conservative about its API, or whole Swift community can benefit from a functional broadcast operator. Even if a major version would remove it, it would still be better to have it for some time.

jacobsapps:
Has support for this package gone the way of Combine?

If merging this is going to be a long way away, has anyone already made a fork that introduces this? If not I will do it

dylanjrafferty:
Just noting that #110 has been recently active in discussing this issue as well

Dev1an:
I'm sure a lot of work is happening that is not always visible towards everyone. @parkera could you maybe list the open questions around sequences and escapability that first need resolution in the standard library before people can start working on this subject again?

parkera:
@phausler is looking into this

Sajjon:
P Hausler wrote a proposal of "share" algorithm yesterday:

https://forums.swift.org/t/kickoff-of-a-new-season-of-development-for-asyncalgorithms-share/81447

PR #215, opened 2022-10-14
[WIP] Deadline algorithm

phausler:
This is a really common need; to create a deadline for given work and throw if that deadline passes.

### DIFF
--- a/Sources/AsyncAlgorithms/Deadline.swift
+++ b/Sources/AsyncAlgorithms/Deadline.swift
@@ -38,3 +38,23 @@ public func withDeadline<C: Clock, T: Sendable>(
     }
   }
 }
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public func withTimeout<C: Clock, T: Sendable>(
+  in duration: C.Duration,
+  clock: C,
+  _ operation: @Sendable () async throws -> T
+) async throws -> T {
+  try await withDeadline(
+    clock.now.advanced(by: duration),
+    clock: clock,
+    operation)
+}
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public func withTimeout<T: Sendable>(
+  in duration: Duration,
+  _ operation: @Sendable () async throws -> T
+) async throws -> T {
+  try await withTimeout(in: duration, clock: .continuous, operation)
+}


FranzBusch (commented):
I haven’t even seen this yet. I am a bit reluctant on this since it is called timeout however it doesn’t really do that since task cancellation is cooperative. The only real way to implement deadlines right now is to use unstructured concurrency.
Though I still agree this has value just unsure if we should call it timeout.

ktoso:
That's a bit of a misleading API... what people ask for and reinvent constantly currently is the unstructured task based one; this likely isn't what people expect by looking at the name.

We could tackle this by using a more descriptive name here, like `withStructuredDeadline` or something else to suggest that this is a structured concurrency thing, and thus WILL await anyway, but it'll cancel at least.

Sources/AsyncAlgorithms/Deadline.swift
```diff
@@ -0,0 +1,60 @@
...
+      }
+      defer { group.cancelAll() }
+      return try await group.next()!
+    }
+  }
+}
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public func withTimeout<C: Clock, T: Sendable>(
+  in duration: C.Duration,
```
ktoso, line 44: I don't think such should be called "timeout", rather this still forms _a deadline_ and should be called `with...Deadline(in: .seconds...)` because it is defining a _deadline_ _in_ some seconds, it is not a timeout "in".

Sources/AsyncAlgorithms/Deadline.swift
```diff
@@ -0,0 +1,60 @@
...
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+public struct TimeoutError<C: Clock>: Error {
```
ktoso, line 13: For consistency this likely should be a `DeadlineExceededError`, we don't speak in terms of timeouts; timeout is the convenience API

Sources/AsyncAlgorithms/Deadline.swift
phausler: isn't a deadline in 5 seconds just a timeout starting at now in 5 seconds?

Sources/AsyncAlgorithms/Deadline.swift
phausler: Yea that might be a better name.

Sources/AsyncAlgorithms/Deadline.swift
FranzBusch: I agree here with @ktoso that we should define this with the term `deadline`. The difference between a deadline and a timeout is quite important. A deadline can be reused and it will always be the same point in time w.r.t. the clock used. A timeout would need to recalculated against your current time. Furthermore, a deadline is less error prone since suspended tasks/threads won't influence the actual point in time calculation.

FranzBusch:
> Per the structured versus unstructured case: it only means that the deadline passing and the work child task being cancelled only becomes an issue if that work child task is unresponsive for cancellation. The cases where this is an issue is where that child task is actually a function that does not participate in cooperative cancellation; e.g. it does not fully participate in swift concurrency. That all being said; I am not sure that a withDeadline that is structured would really be that great in that it will likely end up orphaning tasks when used.

In general, I am with you here and I agree that we should offer this functionality; however, I think we really need to call out that this expects cooperative cancellation and might exceed the deadline. You never really know if some method that you call is cancelling at the first instance that cancellation is raised.

Fab1n:
This depends on iOS 16 and such, just because the dependency on the new clock feature. Would love to see a backport compatible version for iOS 13 and up. 
What do you think?

Keithbreadley (approved):

ph1ps:
@phausler 

Since you've announced that version 1.1 is on track again, I wanted to bring this algorithm to attention. I've already seen tenths of versions of this floating around in the eco system (and I am currently maintaining an open source version of this myself), and it'd be really nice if this could get integrated into AsyncAlgorithms. 

Please let me know if I can be of any help.

Thank you.

phausler:
for the version you are maintaining, what alterations have been needed to be made if any?

ph1ps:
> This is handled in my method with nonisolated(nonsending) which is the latest spelling for this

> The with prefix is aligned with the rest of the ecosystem e.g. withTaskGroup etc.

Right, I should've mentioned that I am not pushing towards the interface I came up with, I am sure there are better ways. I was just explaining the differences from my implementation to the implementation in the MR.

ph1ps:
@FranzBusch Cool. 

In your previous post, you mentioned the ideal solution would be non-@isolated(any), why is this desirable? If closure isolation would land you could eliminate the hop in `addTask`, by doing `addTask { [isolated closure.isolation] in ... }`, no?

Also, I remembered my reasoning behind not going with the with-prefix: in my mind a with-style functions always had an argument that is scoped to the closure, like you mentioned with `withTaskGroup`. `withDeadline` does not have any resource. But this argument does not hold up in current Swift, I guess, because of `withTaskCancellationHandler` and `withTaskPriorityEscalationHandler` which also have no resource being passed along.

ph1ps:
> That's right. We probably want @isolated(any) to avoid the hops. I wasn't able to get that to work yet.

This might be either too constraining or maybe also over engineering the problem but I think it would be possible to pass the closure right along `addTask` like this `taskGroup.addTask(operation: operation)`. The isolation should be retained, as far as I understood the relevant portions in stdlib.

You will lose the possibility to use your `TaskResult` but I think you could use "error-based" state management instead. If `Task.sleep` throws, you can catch the error and rethrow a special `SleepWasCancelledError` that can then be caught in the `withTaskGroup` closure so you can finally call `taskGroup.next()` again to let the closure propagate its cancellation behavior.

To get around `@escaping`, someone would have to take a look at https://github.com/swiftlang/swift/issues/76604, though. I had a look at it a while back, but it seemed to be out of my league.

PR #219, opened 2022-10-19
Remove Sendability from all Iterators

FranzBusch:
# Motivation
We have been discussing this a lot in the other PRs, but in general said that an `AsyncIterator` is the connection between the consuming task and the `AsyncSequence`. If an `AsyncIterator` would be `Sendable` then we don't have that relationship anymore. Therefore, no iterator of an `AsyncIterator` should be `Sendable`.

# Modification
Remove all `Sendable` conformances of the iterators even the conditional ones. We can always add them back later.


phausler (approved):
Looks good to me.

phausler:
One follow on task; we should update the Sources/AsyncAlgorithms/AsyncAlgorithms.docc/Guides/Effects.md file to reflect this.

FranzBusch:
@phausler Good catch! Updated the doc. Feel free to merge the PR

ktoso:
🎉

PR #220, opened 2022-10-20
Remove Sendable from `AsyncBufferedByteIterator`

FranzBusch:
# Motivation
I raise on the pitch of `AsyncBufferedByteIterator` on the forums that I think the iterator must not be `Sendable`. The reasoning for this is twofold. First, an iterator is the connection of a consumer to the `AsyncSequence`; therefore, iterators should not be shared since it breaks that assumption. Secondly, the implementation of the `AsyncBufferedByteIterator` can be more straight forward since we don't have to check for unique ownership of the storage.

# Modification
Remove the `Sendable` conformances from  `AsyncBufferedByteIterator`.

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncBufferedByteIterator.swift
+++ b/Sources/AsyncAlgorithms/AsyncBufferedByteIterator.swift
@@ -39,7 +39,7 @@
 ///     }
 ///
 ///
-public struct AsyncBufferedByteIterator: AsyncIteratorProtocol, Sendable {
+public struct AsyncBufferedByteIterator: AsyncIteratorProtocol {
   public typealias Element = UInt8
   @usableFromInline var buffer: _AsyncBytesBuffer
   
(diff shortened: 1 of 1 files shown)


Sources/AsyncAlgorithms/AsyncBufferedByteIterator.swift
```diff
@@ -110,24 +110,6 @@ internal struct _AsyncBytesBuffer: @unchecked Sendable {
...
-      // edge case of making the iterator and sending that across to two
-      // places to iterate is asking for something bizzare and the answer
-      // should not be crash, but it definitely cannot be consistent.
-      //
-      // The unique ref check is here to prevent the potentials of a crashing
-      // scenario.
-      if !isKnownUniquelyReferenced(&storage) {
-        // The count is not mutated across invocations so the access is safe.
-        let capacity = storage.buffer.count
-        storage = Storage(capacity: capacity)
```
phausler, line 129: if this is being removed then two things: we should definitely in this case make the type marked as explicitly non-sendable via the unavailable annotation, and this then makes the `var storage` then no longer needed as a var but instead can be a let.

phausler (commented):
given the two issues raised I think this will be fine.

Sources/AsyncAlgorithms/AsyncBufferedByteIterator.swift
FranzBusch: Good points! Updated the PR

phausler (approved):
Looks great!

PR #222, opened 2022-10-20
Deferred algorithm

tcldr:
# Deferred

* Proposal: [NNNN](NNNN-deferred.md)
* Authors: [Tristan Celder](https://github.com/tcldr)
* Review Manager: TBD
* Status: **Awaiting feedback**

* Implementation: [[Source](https://github.com/tcldr/swift-async-algorithms/blob/pr/deferred/Sources/AsyncAlgorithms/AsyncDeferredSequence.swift) | 
[Tests](https://github.com/tcldr/swift-async-algorithms/blob/pr/deferred/Tests/AsyncAlgorithmsTests/TestDeferred.swift)]
* Decision Notes: [Additional Commentary](https://forums.swift.org/)
* Bugs:

## Introduction

`AsyncDeferredSequence` provides a convenient way to postpone the initialization of a sequence to the point where it is requested by a sequence consumer. 

## Motivation

Some source sequences may perform expensive work on initialization. This could be network activity, sensor activity, or anything else that consumes system resources. While this can be mitigated in some simple situtations by only passing around a sequence at the point of use, often it is favorable to be able to pass a sequence to its eventual point of use without commencing its initialization process. This is especially true for sequences which are intended for multicast/broadcast for which a reliable startup and shutdown procedure is essential.

A simple example of a seqeunce which may benefit from being deferred is provided in the documentation for AsyncStream:

```swift
extension QuakeMonitor {

    static var quakes: AsyncStream<Quake> {
        AsyncStream { continuation in
            let monitor = QuakeMonitor()
            monitor.quakeHandler = { quake in
                continuation.yield(quake)
            }
            continuation.onTermination = { @Sendable _ in
                 monitor.stopMonitoring()
            }
            monitor.startMonitoring()
        }
    }
}
```

In the supplied code sample, the closure provided to the AsyncStream initializer will be executed immediately upon initialization; `QuakeMonitor.startMonitoring()` will be called, and the stream will then begin buffering its contents waiting to be iterated. Whilst this behavior is sometimes desirable, on other occasions it can cause system resources to be consumed unnecessarily.

```swift
let nonDeferredSequence = QuakeMonitor.quakes //  `Quake.startMonitoring()` is called now!

...
// at some arbitrary point, possibly hours later...
for await quake in nonDeferredSequence {
    print("Quake: \(quake.date)")
}
// Prints out hours of previously buffered quake data before showing the latest
```

## Proposed solution

`AsyncDeferredSequence` provides a way to postpone the initialization of an an arbitrary async sequence until the point of use:

```swift
let deferredSequence = deferred(QuakeMonitor.quakes) // Now, initialization is postponed

...
// at some arbitrary point, possibly hours later...
for await quake in deferredSequence {  //  `Quake.startMonitoring()` is now called
    print("Quake: \(quake.date)")
}
// Prints out only the latest quake data
```

Now, potentially expensive system resources are consumed only at the point they're needed.

## Detailed design

`AsyncDeferredSequence` is a trivial algorithm supported by some convenience functions.

### Functions

```swift
public func deferred<Base: AsyncSequence>(_ createSequence: @escaping @Sendable () async -> Base) -> AsyncDeferredSequence<Base>
public func deferred<Base: AsyncSequence>(_ createSequence: @autoclosure @escaping @Sendable () -> Base) -> AsyncDeferredSequence<Base>
```

The synchronous function can be auto-escaped, simplifying the call-site. While the async variant allows a sequence to be initialized within a concurrency context other than that of the end consumer.

```swift
public struct AsyncDeferredSequence<Base: AsyncSequence>: Sendable {
  public typealias Element = Base.Element
  public struct Iterator: AsyncIteratorProtocol {
    public mutating func next() async rethrows -> Element?
  }
  public func makeAsyncIterator() -> Iterator
}
```

### Naming

The `deferred(_:)` function takes its inspiration from the Combine publisher of the same name with similar functionality. However, `lazy(_:)` could be quite fitting, too.

### Comparison with other libraries

**ReactiveX** ReactiveX has an [API definition of Defer](https://reactivex.io/documentation/operators/defer.html) as a top level operator for generating observables.

**Combine** Combine has an [API definition of Deferred](https://developer.apple.com/documentation/combine/deferred) as a top-level convenience publisher.

## Effect on API resilience

Deferred has a trivial implementation and is marked as `@frozen` and `@inlinable`. This removes the ability of this type and functions to be ABI resilient boundaries at the benefit of being highly optimizable.

### DIFF
--- a/Evolution/NNNN-deferred.md
+++ b/Evolution/NNNN-deferred.md
@@ -1,0 +1,109 @@
+# Deferred
+
+* Proposal: [NNNN](NNNN-deferred.md)
+* Authors: [Tristan Celder](https://github.com/tcldr)
+* Review Manager: TBD
+* Status: **Awaiting implementation**
+
+* Implementation: [[Source](https://github.com/tcldr/swift-async-algorithms/blob/pr/deferred/Sources/AsyncAlgorithms/AsyncDeferredSequence.swift) | 
+[Tests](https://github.com/tcldr/swift-async-algorithms/blob/pr/deferred/Tests/AsyncAlgorithmsTests/TestDeferred.swift)]
+* Decision Notes: [Additional Commentary](https://forums.swift.org/)
+* Bugs:
+
+## Introduction
+
+`AsyncDeferredSequence` provides a convenient way to postpone the initialization of a sequence to the point where it is requested by a sequence consumer. 
+
+## Motivation
+
+Some source sequences may perform expensive work on initialization. This could be network activity, sensor activity, or anything else that consumes system resources. While this can be mitigated in some simple situtations by only passing around a sequence at the point of use, often it is favorable to be able to pass a sequence to its eventual point of use without commencing its initialization process. This is especially true for sequences which are intended for multicast/broadcast for which a reliable startup and shutdown procedure is essential.
+
+A simple example of a seqeunce which may benefit from being deferred is provided in the documentation for AsyncStream:
+
+```swift
+extension QuakeMonitor {
+
+    static var quakes: AsyncStream<Quake> {
+        AsyncStream { continuation in
+            let monitor = QuakeMonitor()
+            monitor.quakeHandler = { quake in
+                continuation.yield(quake)
+            }
+            continuation.onTermination = { @Sendable _ in
+                 monitor.stopMonitoring()
+            }
+            monitor.startMonitoring()
+        }
+    }
+}
+```
+
+In the supplied code sample, the closure provided to the AsyncStream initializer will be executed immediately upon initialization; `QuakeMonitor.startMonitoring()` will be called, and the stream will then begin buffering its contents waiting to be iterated. Whilst this behavior is sometimes desirable, on other occasions it can cause system resources to be consumed unnecessarily.
+
+```swift
+let nonDeferredSequence = QuakeMonitor.quakes //  `Quake.startMonitoring()` is called now!
+
+...
+// at some arbitrary point, possibly hours later...
+for await quake in nonDeferredSequence {
+    print("Quake: \(quake.date)")
+}
+// Prints out hours of previously buffered quake data before showing the latest
+```
+
+## Proposed solution
+
+`AsyncDeferredSequence` provides a way to postpone the initialization of an an arbitrary async sequence until the point of use:
+
+```swift
+let deferredSequence = deferred(QuakeMonitor.quakes) // Now, initialization is postponed
+
+...
+// at some arbitrary point, possibly hours later...
+for await quake in deferredSequence {  //  `Quake.startMonitoring()` is now called
+    print("Quake: \(quake.date)")
+}
+// Prints out only the latest quake data
+```
+
+Now, potentially expensive system resources are consumed only at the point they're needed.
+
+## Detailed design
+
+`AsyncDeferredSequence` is a trivial algorithm supported by some convenience functions.
+
+### Functions
+
+```swift
+public func deferred<Base: AsyncSequence>(_ createSequence: @escaping @Sendable () async -> Base) -> AsyncDeferredSequence<Base>
+public func deferred<Base: AsyncSequence>(_ createSequence: @autoclosure @escaping @Sendable () -> Base) -> AsyncDeferredSequence<Base>
+```
+
+The synchronous function can be auto-escaped, simplifying the call-site. While the async variant allows a sequence to be initialized within a concurrency context other than that of the end consumer.
+
+```swift
+public struct AsyncDeferredSequence<Base: AsyncSequence>: Sendable {
+  public typealias Element = Base.Element
+  public struct Iterator: AsyncIteratorProtocol {
+    public mutating func next() async rethrows -> Element?
+  }
+  public func makeAsyncIterator() -> Iterator
+}
+```
+
+### Naming
+
+The `deferred(_:)` function takes its inspiration from the Combine publisher of the same name with similar functionality. However, `lazy(_:)` could be quite fitting, too.
+
+### Comparison with other libraries
+
+**ReactiveX** ReactiveX has an [API definition of Defer](https://reactivex.io/documentation/operators/defer.html) as a top level operator for generating observables.
+
+**Combine** Combine has an [API definition of Deferred](https://developer.apple.com/documentation/combine/deferred) as a top-level convenience publisher.
+
+
+## Effect on API resilience
+
+Deferred has a trivial implementation and is marked as `@frozen` and `@inlinable`. This removes the ability of this type and functions to be ABI resilient boundaries at the benefit of being highly optimizable.
+
+## Alternatives considered
--- a/Sources/AsyncAlgorithms/AsyncDeferredSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncDeferredSequence.swift
@@ -1,0 +1,77 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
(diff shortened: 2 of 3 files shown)


Evolution/NNNN-deferred.md
```diff
@@ -0,0 +1,109 @@
...
+
+```swift
+public func deferred<Base: AsyncSequence>(_ createSequence: @escaping @Sendable () async -> Base) -> AsyncDeferredSequence<Base>
+public func deferred<Base: AsyncSequence>(_ createSequence: @autoclosure @escaping @Sendable () -> Base) -> AsyncDeferredSequence<Base>
+```
+
+The synchronous function can be auto-escaped, simplifying the call-site. While the async variant allows a sequence to be initialized within a concurrency context other than that of the end consumer.
+
+```swift
+public struct AsyncDeferredSequence<Base: AsyncSequence>: Sendable {
```
phausler, line 85: I am going to guess that the `Base` MUST be `Sendable`

Tests/AsyncAlgorithmsTests/TestDeferred.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+    XCTAssertEqual(expected, actual)
+  }
+  
+  func test_deferred_remains_idle_pending_consumer() async {
+    let expectation = expectation(description: "pending")
+    expectation.isInverted = true
+    let _ = deferred {
+      AsyncStream { continuation in
+        expectation.fulfill()
```
phausler, line 33: Sneaky use of expectations! <3

Tests/AsyncAlgorithmsTests/TestDeferred.swift
```diff
@@ -0,0 +1,82 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+@preconcurrency import XCTest
+import AsyncAlgorithms
+
+final class TestDeferred: XCTestCase {
```
phausler, line 15: For completeness we should probably have tests for cancellation and iterating past the end.

Sources/AsyncAlgorithms/AsyncDeferredSequence.swift
```diff
@@ -0,0 +1,77 @@
...
+      case .pending(let generator):
+        state = .active(await generator().makeAsyncIterator())
+        return try await next()
+      case .active(var base):
+        if let value = try await base.next() {
+          state = .active(base)
+          return value
+        }
+        else {
+          return nil
```
phausler, line 65: we could enforce a `.terminal` state here too and if the base throws or returns nil it puts it into that `.terminal` which would return nil immediately upon call of next.

Sources/AsyncAlgorithms/AsyncDeferredSequence.swift
```diff
@@ -0,0 +1,77 @@
...
+public func deferred<Base: AsyncSequence>(_ createSequence: @escaping @Sendable () async -> Base) -> AsyncDeferredSequence<Base> {
+  AsyncDeferredSequence(createSequence)
+}
+
+@inlinable
+public func deferred<Base: AsyncSequence>(_ createSequence: @autoclosure @escaping @Sendable () -> Base) -> AsyncDeferredSequence<Base> {
+  AsyncDeferredSequence(createSequence)
+}
+
+public struct AsyncDeferredSequence<Base: AsyncSequence> {
```
phausler, line 22: I think this one might benefit from being frozen.

Sources/AsyncAlgorithms/AsyncDeferredSequence.swift
```diff
@@ -0,0 +1,77 @@
...
+  
+  @usableFromInline
+  let createSequence: @Sendable () async -> Base
+  
+  @usableFromInline
```
phausler, line 27: if it is frozen the initializers should then be `@inlinable`. (including the iterator too)

Sources/AsyncAlgorithms/AsyncDeferredSequence.swift
tcldr: Ah, yes. Of course! Release the iterator, too. Thanks!

tcldr:
Great – thanks for the feedback! Should be incorporated now.

Sources/AsyncAlgorithms/AsyncDeferredSequence.swift
```diff
@@ -0,0 +1,92 @@
...
+
+@inlinable
+public func deferred<Base: AsyncSequence & Sendable>(
+  _ createSequence: @autoclosure @escaping @Sendable () -> Base
+) -> AsyncDeferredSequence<Base> where Base: AsyncSequence, Base: Sendable {
+  AsyncDeferredSequence(createSequence)
+}
+
+@frozen
+public struct AsyncDeferredSequence<Base> where Base: AsyncSequence, Base: Sendable {
```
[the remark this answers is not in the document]
tcldr: I've had a go – hopefully its clear!

Evolution/NNNN-deferred.md
```diff
@@ -0,0 +1,114 @@
...
+
+...
+// at some arbitrary point, possibly hours later...
+for await quake in nonDeferredSequence {
+    print("Quake: \(quake.date)")
+}
+// Prints out hours of previously buffered quake data before showing the latest
+```
+
+## Proposed solution
```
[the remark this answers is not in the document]
tcldr: Yes, good point. I've tried to clarify.

Evolution/NNNN-deferred.md
```diff
@@ -0,0 +1,114 @@
...
+
+`AsyncDeferredSequence` uses a supplied closure to create a new asynchronous sequence each time it is iterated. This has the effect of postponing the initialization of an an arbitrary async sequence until the point of use:
```
FranzBusch, line 56:
```suggestion
`AsyncDeferredSequence` uses a supplied closure to create a new asynchronous sequence. The closure is executed for each iterator on the first call to `next`. This has the effect of postponing the initialization of an arbitrary async sequence until the point of first demand:
```

Sources/AsyncAlgorithms/AsyncDeferredSequence.swift
```diff
@@ -0,0 +1,99 @@
...
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+/// Creates an asynchronous sequence that uses the supplied closure to create a new asynchronous sequence each
+/// time it is iterated
```
FranzBusch, line 13:
```suggestion
/// Creates a ``AsyncDeferredSequence`` that uses the supplied closure to create a new `AsyncSequence`.
/// The closure is executed for each iterator on the first call to `next`.
/// This has the effect of postponing the initialization of an arbitrary `AsyncSequence` until the point of first demand.
```

Sources/AsyncAlgorithms/AsyncDeferredSequence.swift
```diff
@@ -0,0 +1,99 @@
...
+@inlinable
+public func deferred<Base>(
+  _ createSequence: @escaping @Sendable () async -> Base
+) -> AsyncDeferredSequence<Base> where Base: AsyncSequence, Base: Sendable {
+  AsyncDeferredSequence(createSequence)
+}
+
+/// Creates an asynchronous sequence that uses the supplied closure to create a new asynchronous sequence each
+/// time it is iterated
```
FranzBusch, line 22:
```suggestion
/// Creates a ``AsyncDeferredSequence`` that uses the supplied closure to create a new `AsyncSequence`.
/// The closure is executed for each iterator on the first call to `next`.
/// This has the effect of postponing the initialization of an arbitrary `AsyncSequence` until the point of first demand.
```

tcldr:
@FranzBusch yes, much clearer. Thanks!

Keithbreadley (approved):

PR #228, opened 2022-11-07
Rework `combineLatest` to remove the `Sendable` constraint on the base iterator

FranzBusch:
# Motivation
This follows the recent rework of the `merge` and `zip` algorithms where we dropped the `Sendable` constraint on the `BaseX.AsyncIterator`. This PR now removes the same constraint on `combineLatest`.

# Modification
This PR overhauls the implementation of `combineLatest` with the goal of changing these things:
- Drop the `Sendable` constraint of the base `AsyncIterator`s
- Only create a single `Task`

# Result
`combineLatest` now doesn't require the base `AsyncIterator` to be `Sendable`


Tests/AsyncAlgorithmsTests/TestCombineLatest.swift
```diff
@@ -19,7 +19,6 @@ final class TestCombineLatest2: XCTestCase {
     let sequence = combineLatest(a.async, b.async)
     let actual = await Array(sequence)
     XCTAssertGreaterThanOrEqual(actual.count, 3)
-    XCTAssertEqual(actual.first!, (1, "a"))
```
phausler, line 22: are we no longer guaranteed that the first element is effectively a zip?

twittemb: Hi

Do you know what’s the behaviour for similar operators in other libs like Combine and Rx… ?

Couldn’t we suspend the first call to next until both elements are available ?

FranzBusch: This tests does hit a special case here since it has two bases that have values buffered, in a normal continuous flow of values we will see the regular combining pattern.

The thing is we can't really do any timing here. What we are doing is in both child tasks to request from the upstreams. We cannot suspend the first child task until the second produces something because that might happen in X minutes and we need to continue to consume all elements here until the other produced one. Combine has the same behaviour, but you can't compare it really because Combine is push based where as `AsyncSequence`s are pulled based.

twittemb: In the following case

let a = [1, 2, 3]
let b = ["a"]
let sequence = combineLatest(a.async, b.async)

Does it mean that the result could be one of the following depending on the scheduling of tasks?
(1, “a”) (2, “a”) (3, “a”)
Or
(2, “a”) (3, “a”)
Or
(3, “a”)

I would expect only the first one.

what’s the issue with suspending the first next (even for a long period of time) waiting for the first other element?

Is there like an “official” definition of what combineLatest should be ?
I feel like the implementation is a mix between combineLatest and withLatestFrom.

FranzBusch: You are correct here that any of the three outputs is valid. There is no guarantee how the child tasks for each upstream are scheduled. Imagine a system where the cooperative thread pool consists of only a single thread. That means we can only pull from one upstream at a time.

> what’s the issue with suspending the first next (even for a long period of time) waiting for the first other element?
This goes against the definition of `combineLatest`. The first element is produced when all of the upstreams produced one value until that happens we always need to continue to demand values from the ones that already produced one.

You can play around with this over here: https://rxmarbles.com/#combineLatest

As I said above already, the test is pretty special since the bases are sequences. In a real world , where the elements are produced over time this is behaves as expected.

twittemb: Ok makes sens. Thanks for your answer.

phausler: Combine does have the guarantee that the first tuple produced behaves like a zip, and then after that it updates the value to the latest and produces a value (upon demand).

FranzBusch: Are you sure. The following Combine example produces the same output
```swift
let a = [1, 2, 3].publisher
let b = ["a", "b", "c"].publisher
let c = a.combineLatest(b)
    .sink { print($0) }

// (3, "a")
// (3, "b")
// (3, "c")
```

Which is IMO correct. Until all upstreams produced an element `combineLatest` issues a constant demand to all upstreams.

PR #230, opened 2022-11-16
Cleanup Sendable conformances and @preconcurrency imports to build as strict concurrency clean

phausler:
This corrects a number of stray requirements for `Sendable` on iterators and adjusts a few cases of no-longer-needed `@preconcurrency` annotations. This now brings us to a fully clean `-strict-concurrency=complete` build.


Sources/AsyncAlgorithms/AsyncThrowingChannel.swift
```diff
@@ -9,7 +9,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import OrderedCollections
+@preconcurrency import OrderedCollections
```
phausler, line 12: I wonder if we should mark these as `@_implementationOnly`

FranzBusch: IMO, I consider it a best practice nowadays to mark every import as `@_implementationOnly` it really should be the default.

phausler: I will follow-up with an additional audit of those

FranzBusch (approved):
LGTM!

PR #233, opened 2022-11-17
Transform the reductions and remove duplicates guides into a proposal

phausler:
This proposal formalizes the family of Async*Reductions types and optimized derivatives.

Read the full proposal [here](https://github.com/phausler/swift-async-algorithms/blob/pr/reductions_proposal/Evolution/NNNN-reductions.md).


KevinD313 (approved):
Yeah

PR #236, opened 2022-11-21
Add a proposal for AsyncBufferSequence

phausler:
This outlines a proposal for the `AsyncBufferSequence` type and the `AsyncBuffer` protocol used for buffering elements produced by an `AsyncSequence`

Read the full proposal [here](https://github.com/phausler/swift-async-algorithms/blob/pr/buffer_proposal/Evolution/NNNN-buffer.md)

### DIFF
--- a/Evolution/NNNN-buffer.md
+++ b/Evolution/NNNN-buffer.md
@@ -1,0 +1,172 @@
+# Buffer
+
+* Proposal: [SAA-NNNN](https://github.com/apple/swift-async-algorithms/blob/main/Evolution/NNNN-buffer.md)
+* Author(s): [Philippe Hausler](https://github.com/phausler)
+* Status: **Implemented**
+* Implementation: [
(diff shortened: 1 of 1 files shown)


Keithbreadley (approved):

twittemb:
I guess we should update the proposal to reflect the new implementation, right?

phausler:
yup that is definitely something that needs to be done, right now I've got refinements for `Observable` on my docket so once that is done I can get back around to updating it (unless someone else wants to modify and take that on)

twittemb:
> yup that is definitely something that needs to be done, right now I've got refinements for `Observable` on my docket so once that is done I can get back around to updating it (unless someone else wants to modify and take that on)

I can give it a try. That way you will just to do a sanitization pass

twittemb:
I guess this PR can be abandoned since the new buffer algorithm is merged?

FranzBusch:
Yep closing it!

PR #241, opened 2022-12-15
AsyncChannel: Use close verb instead of finish

eaigner:
When we are talking about channels, we usually close them (e.g. same for file descriptors, pipes etc.).

It is also named `close` in Kotlin, Go etc.

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncChannel.swift
+++ b/Sources/AsyncAlgorithms/AsyncChannel.swift
@@ -18,8 +18,8 @@
 /// pressure applied by `send(_:)` via the suspension/resume ensures that
 /// the production of values does not exceed the consumption of values from iteration. This method
 /// suspends after enqueuing the event and is resumed when the next call to `next()`
-/// on the `Iterator` is made, or when `finish()` is called from another Task.
-/// As `finish()` induces a terminal state, there is no need for a back pressure management.
+/// on the `Iterator` is made, or when `close()` is called from another Task.
+/// As `close()` induces a terminal state, there is no need for a back pressure management.
 /// This function does not suspend and will finish all the pending iterations.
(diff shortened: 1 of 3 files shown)


twittemb:
> /cc @twittemb Since I saw you reimplemented `AsyncChannel` in a different PR

Thanks for keeping me in the loop. If my PR is approved I’ll make sure the naming of in sync.

eaigner:
Also /cc @phausler since the proposal seems to originate from you

phausler:
Close has some distinct implications; AsyncStream uses the `finish` terminology and our documentation around `AsyncSequence` in general uses finish too. So I am not fully convinced that this should have a distinct name in comparison to other systems that are related. (yield versus send are named distinctly since they do different things)

eaigner:
Mhm. `AsyncStream.Continuation` is the only async type in Swift that has a `finish` method, and its not actually on the `AsyncStream` itself, but its continuation. On the other hand `close` would be called on the channel, not some continuation.

Screenshot 2022-12-16 at 19 19 44

kelvin13:
in my mind, ‘close’ is bidirectional. either side can *close* a medium.

‘finish’ is directional. the consumer *stops* (or *terminates*) consumption, the producer *finishes* producing values.

i think the word finish has the connotation that it wouldn't make sense to continue communication because there is nothing left to communicate. closing or stopping communication doesn't suggest there is nothing left to communicate.

just my 2c

eaigner:
I guess it would fit the theme of Apple constantly claiming they invented something that existed for ages.

- https://gobyexample.com/closing-channels
- https://kotlinlang.org/docs/channels.html#closing-and-iteration-over-channels
- http://www.di.uevora.pt/~lmr/syscalls.html section "Creating a Channel"

eaigner:
I see this is pointless. No wonder the Swift language and ecosystem is such a dumpster fire.

PR #243, opened 2023-01-03
Add precondition guards to pthread_mutex calls

Danappelxx:
While in the current configuration lock/unlock shouldn't ever fail, it seems polite to guard against basic misuse in debug builds.


phausler (approved):

kperryua:
I think you may want to pull the result of the lock/unlock calls out of the `precondition()`. According to the docs:

>In -Ounchecked builds, condition is not evaluated, but the optimizer may assume that it always evaluates to true. Failure to satisfy that assumption is a serious programming error.

So in `-Ounchecked` builds, there will be no locking at all in these paths.

phausler:
err good call @kperryua; will adjust that

PR #250, opened 2023-02-01
Tolerate cancelling before we create our task

FranzBusch:
# Motivation
We switched the point in time when we create tasks to the first call to `next()`. This had the side effect that we can now be cancelled before we even created our task. Which by itself is fine but we didn't handle it in the state machines.

# Modification
Adapt the state machines of `debounce`, `zip`, `merge` and `combineLatest` to handle cancellation before `next()`.

# Result
We shouldn't run into these preconditions anymore.


phausler (approved):
Good catch!

Issue #253, opened 2023-02-16
Several tests fail or hang on linux with the official trunk snapshot toolchains


finagolfin:
I see this both on linux x86_64 and Android AArch64. Is there no linux CI for this repo?

FranzBusch:
Are you still seeing this issue?

buttaface:
Yes, I see several tests hanging and failing on linux x86_64 with the latest Feb. 21 official snapshot build for Ubuntu 20.04 x86_64 testing the latest trunk commit ffa6ee9. In particular, these tests all hang on Android AArch64 also:
```
AsyncAlgorithmsTests.TestCombineLatest2/test_throwing_ordering1
AsyncAlgorithmsTests.TestCombineLatest2/test_throwing_ordering2
AsyncAlgorithmsTests.TestZip2/test_zip_produces_one_element_and_throws_when_first_produces_one_element_and_throws
AsyncAlgorithmsTests.TestZip2/test_zip_produces_one_element_and_throws_when_second_produces_one_element_and_throws
AsyncAlgorithmsTests.TestZip3/test_zip_produces_one_element_and_throws_when_first_produces_one_element_and_throws
AsyncAlgorithmsTests.TestZip3/test_zip_produces_one_element_and_throws_when_second_produces_one_element_and_throws
AsyncAlgorithmsTests.TestZip3/test_zip_produces_one_element_and_throws_when_third_produces_one_element_and_throws
 ```
A handful more also fail, let me know if you cannot reproduce.

**Comment by FranzBusch** [comment]
This PR should fix it: https://github.com/apple/swift-async-algorithms/pull/254

**Comment by buttaface** [comment]
I'm still seeing a few sporadically hanging tests on Android AArch64, particularly if all the tests are run in parallel. This is when building the latest trunk commit from a couple weeks ago, e6b21d2.

[I just cross-compiled the latest trunk snapshot toolchain for Android AArch64](https://github.com/buttaface/termux-packages/actions/runs/4482777560) and the following test hung once out of 10 runs of `swift test --parallel`:
```
TestBuffer/test_given_a_buffered_bounded_sequence_when_cancelling_consumer_then_the_iteration_finishes_and_the_base_is_cancelled
```
I also tried running 10 times with the last March 16 trunk source snapshot that I built natively on my Android phone, and that previous test hung once and this test hung once:
```
TestZip2/test_zip_finishes_when_iteration_task_is_cancelled
```

FranzBusch:
[...]

PR #254, opened 2023-02-27
Switch from `group.waitForAll()` to `group.next()`

FranzBusch:
# Motivation
Swift 5.8 is including a change to how `group.waitForAll()` is working. It now properly waits for all tasks to finish even if one of the tasks throws. We have used `group.waitForAll()` in multiple places and need to change this code accordingly.

# Modification
Switch code from `group.waitForAll()` to `group.next()`.

# Result
This fixes a few stuck tests that we have seen when running against development snapshots.

### DIFF
--- a/Sources/AsyncAlgorithms/CombineLatest/CombineLatestStateMachine.swift
+++ b/Sources/AsyncAlgorithms/CombineLatest/CombineLatestStateMachine.swift
@@ -530,7 +530,8 @@ struct CombineLatestStateMachine<
       preconditionFailure("Internal inconsistency current state \(self.state) and received upstreamThrew()")
 
     case .upstreamsFinished:
-      preconditionFailure("Internal inconsistency current state \(self.state) and received upstreamThrew()")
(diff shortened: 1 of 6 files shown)


buttaface:
I can confirm this fixes all the test issues alluded to in #253 on Android AArch64. On linux x86_64, maybe it's because I'm running on a single-core VPS, but I still see issues. If you don't see any problems on a beefier linux box, probably unrelated and you can go ahead with this.

phausler (approved):
If we are sure that the `isEmpty` does not introduce races (which from our side doesn't seem to have any) this looks fine to me.

phausler:
@swift-ci test

ktoso (approved):
This is LGTM and I clarified semantics of the waitForAll now in Swift itself: https://github.com/apple/swift/pull/63956

The implementation actually is pretty much the same in Swift, a loop over isEmpty so this will be ok -- it is not racy since only the parent task can be adding tasks 👍

PR #255, opened 2023-02-28
Package.swift: make package name consistent with repo name

MaxDesiatov:
Rest of SwiftPM packages provided by Apple follow a dash-case naming scheme consistent with repository name, usually 
with `swift-` prefix. See attached screenshot:

Screenshot 2023-02-28 at 12 57 06

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,7 +3,7 @@
 import PackageDescription
 
 let package = Package(
-  name: "AsyncAlgorithms",
+  name: "swift-async-algorithms",
   platforms: [
     .macOS("10.15"),
     .iOS("13.0"),


FranzBusch (approved):

FranzBusch:
Thanks @MaxDesiatov . We should definitely align this before we do a `1.0.0`

twittemb:
We should probably adapt the installation instructions in the README then.

MaxDesiatov:
The name of the package in `README.md` is already specified according to the convention, it was `Package.swift` that wasn't matching these instructions:

```
    .product(name: "AsyncAlgorithms", package: "swift-async-algorithms"),
```

twittemb:
> The name of the package in `README.md` is already specified according to the convention, it was `Package.swift` that wasn't matching these instructions:
> 
> ```
>     .product(name: "AsyncAlgorithms", package: "swift-async-algorithms"),
> ```

yep sorry :-) my bad !

PR #261, opened 2023-04-05
Implement AsyncSequence.enumerated()

adam-fowler:
Pretty much the same as `Sequence.enumerated` but for `AsyncSequence`. 

eg Translates `["a", "b", "c"]` to `[(0,"a"), (1, "b"), (2, "c")]`

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncEnumeratedSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncEnumeratedSequence.swift
@@ -1,0 +1,65 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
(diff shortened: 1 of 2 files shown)


FranzBusch:
Thanks for opening this @adam-fowler! Definitely a needed `AsyncSequence`. We are still finalising the 1.0.0 release. I will mark this as a 1.1.0 candidate

phausler:
I know this has been a very long while since this was posted; however since we are at development for 1.1 this is a really useful algorithm for composing buffer dropping detection and other such things. I would love to run this in the near future - if you are still interested in pushing this forward I would like to see if this is something you would still like to do. Alternatively I can update this and drive it forward if not (still of course giving credit for the original drafts)

phausler:
That probably aligns just right with some of the other algorithms. Let me know when you feel things are ready to go.

I think there are a few implementation details that need to be addressed but the major issue is getting a proposal written.

PR #263, opened 2023-04-07
Rename `downStream` to `downstream`

s2mr:
Sorry for the minor point.
I am curious and will submit a PR.


phausler (approved):
This seems to only affect internal surface areas. Seems reasonable to me.

PR #267, opened 2023-06-01
[AsyncInterspersedSequence] Integrate review feedback

FranzBusch:
This integrates all of the feedback from the review thread. Here is a quick summary:

- Change the trailing separator behaviour. We are no longer returning a separator before we are forwarding the error
- Add a synchronous and asynchronous closure based `interspersed` method.
- Support interspersing every n elements


phausler (approved):
looks good to me

PR #274, opened 2023-06-23
Reduce the overall warnings from tests

phausler:
This resolves almost all of the warnings in the tests (except one).


FranzBusch (approved):

kabiroberai:
:wave: hey @phausler, you might wanna have a look at https://github.com/apple/swift/pull/66934 which should fix the deprecation warning (given that there's not much you can do about it pre-stdlib-5.9). That said, it's worth noting that, iiuc, the `ExecutorJob` overload here will never be used since the `UnownedJob` variant is preferred. If my PR lands, it should hopefully be possible for you to remove the former until pre-5.9 platforms are dropped — or maybe gate it behind a `!canImport(Darwin)` conditional.

PR #281, opened 2023-07-27
Fix more `Sendable` warnings and flaky test

FranzBusch:
# Motivation
We still had some `Sendable` warnings left under strict Concurrency checking.

# Modification
This PR fixes a bunch of `Sendable` warnings but we still have some left in the validation tests. Additionally, I fixed a flaky test.

### DIFF
--- a/Package@swift-5.7.swift
+++ b/Package@swift-5.7.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.8
+// swift-tools-version: 5.6
 
 import PackageDescription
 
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -20,18 +20,33 @@ let package = Package(
...
+      swiftSettings: [
+          .enableExperimentalFeature("StrictConcurrency=complete"),
+      ]
     ),
     .target(
       name: "AsyncSequenceValidation",
-      dependencies: ["_CAsyncSequenceValidationSupport", "AsyncAlgorithms"]),
+      dependencies: ["_CAsyncSequenceValidationSupport", "AsyncAlgorithms"],
+      swiftSettings: [
+          .enableExperimentalFeature("StrictConcurrency=complete"),
```
phausler, line 32: ❤️ perfection!

Package@swift-5.7.swift
```diff
@@ -0,0 +1,51 @@
+// swift-tools-version: 5.6
```
phausler, line 1: should this be 5.7 here?

Package@swift-5.6.swift
```diff
@@ -0,0 +1,51 @@
+// swift-tools-version: 5.6
```
phausler, line 1: The `@swift-5.6` suffix is canonical?

Package@swift-5.7.swift
FranzBusch: This can be 5.6 since we don’t use any 5.7 toolchain feature in here. The name of the file is the important part.

Package@swift-5.6.swift
FranzBusch: What do you mean with canonical? SPM used these suffixes to identify what manifest to pick for a given swift version

phausler: ah, yea I did not realize that swiftly uses those suffixes, seems reasonable then

PR #287, opened 2023-08-16
Remove the validation packages from public products

FranzBusch:
# Motivation
The current validation testing library has a bunch of warnings and we haven't put it through a proper API review yet. Since we are preparing for a 1.0.0, I think it is best if we remove the products for the validation targets for now and take some time to make sure the interfaces are taken through an API review.

# Modification
Remove products for the validation targets.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,9 +12,6 @@ let package = Package(
   ],
   products: [
     .library(name: "AsyncAlgorithms", targets: ["AsyncAlgorithms"]),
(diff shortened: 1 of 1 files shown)


FranzBusch:
@swift-ci please test

phausler:
Ive been considering this one for a bit; and as much as Id love to have validation to be useful outside of the package - I tend to agree that this is a bit out of scope of the package as a 1.0.

CrownedPhoenix:
Has there been any traction on this front?
I would really like to have access to AsyncSequenceValidation to do my own testing of custom stream operators.

PR #288, opened 2023-08-18
Rework availability for executor to avoid warnings

phausler:
This works-around the warnings for availability for the ExecutorJob while retaining the existing behaviors.


phausler:
@swift-ci please test

FranzBusch (approved):

PR #289, opened 2023-08-25
Remove background correction in `AsyncTimerSequence`

FranzBusch:
# Motivation
Currently, the `AsyncTimerSequence` is trying to correct for when an application becomes suspended and the timer might fire multiple times once the application gets foregrounded again.

However, this is already handled by the `Clock` types themselves. The `SuspendingClock` is correcting for suspension of the app whereas the `ContinuousClock` is not.

Additionally, this was not only hit by background an application but by just calling `Task.sleep` in the for-await loop that is consuming the sequence.

# Modification
This removes the part of the code in `AsyncTimerSequence` which corrected for suspension of the application.


phausler (approved):

FranzBusch:
@swift-ci please test

FranzBusch:
@swift-ci please test

PR #290, opened 2023-09-06
Refactor Throttle to use the same engine style as Debounce

phausler:
This re-works throttle to use the same style of state machine as debounce.

Note: all existing tests pass...


phausler:
@swift-ci please test

phausler:
after looking further at this - it is a 100% isomorphism to the existing code...

FranzBusch:
@phausler this can be closed for now right?

PR #291, opened 2023-09-14
make AsyncRemoveDuplicatesSequence's init public

yermukhanbet:
Currently our team is using removeDuplicates() function for the AsyncPublisher in one of our service protocols as a return type. Since for each service we have both live and mock structs, I faced an issue with mocking ASyncRemoveDuplicatesSequence value, as the init is internal.

In this PR I changed it to public. Please, let me know if internal init was intentional approach, so I will try to find another possible way for mocking the return type.


yermukhanbet:
@swift-ci please test

FranzBusch:
Right now this is intentional. We might want to revisit this on a larger scale if we do it and make all `init`s public. cc @phausler

PR #292, opened 2023-09-15
Ensure the last element of reduction in the throttle is emitted and use appropriate delay

phausler:
This corrects and validates the case where the last element of a throttle is within the period of the throttle interval but no subsequent elements are produced after that terminal event.

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncThrottleSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncThrottleSequence.swift
@@ -81,7 +81,16 @@ extension AsyncThrottleSequence: AsyncSequence {
       let start = last ?? clock.now
       repeat {
         guard let element = try await base.next() else {
-          return nil
+          if reduced != nil, let last {
+            // ensure the rate of elements never exceeds the given interval
+            let amount = interval - last.duration(to: clock.now)
+            if amount > .zero {
+              try? await clock.sleep(for: amount)
+            }
+          }
+          // the last value is unable to have any subsequent
+          // values so always return the last reduction
+          return reduced
         }
         let reduction = await reducing(reduced, element)
         let now = clock.now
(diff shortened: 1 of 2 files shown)


phausler:
@swift-ci please test

Sources/AsyncAlgorithms/AsyncThrottleSequence.swift
```diff
@@ -81,7 +81,16 @@ extension AsyncThrottleSequence: AsyncSequence {
       let start = last ?? clock.now
       repeat {
         guard let element = try await base.next() else {
-          return nil
+          if reduced != nil, let last {
+            // ensure the rate of elements never exceeds the given interval
+            let amount = interval - last.duration(to: clock.now)
+            if amount > .zero {
+              try? await clock.sleep(for: amount)
```
FranzBusch, line 88: Why are we swallowing the cancellation error here?

FranzBusch: Wondering if we should throw in the case where we are a throwing sequence and if not just return `nil`.

phausler: We can't really know that - plus throwing `CancellationError` seems rather off to me except in the cases where EOF and cancellation need disambiguation

FranzBusch: It is impossible to differentiate when a non-throwing async sequence returns `nil` if it was due to EOF or task cancellation, that's why I am leaning towards always throwing the `CancellationError` if possible. Now this case is interesting because we would just return the reduced value and probably the next iteration will potentially throw the `CancellationError`. So I guess it's fine

FranzBusch (approved):

Issue #248, opened 2023-01-27
A throttled `AsyncSequence` doesn't yield its parent sequence's final value


gregcotten:
Not sure if this is expected behavior.

I make an `AsyncStream` called `stream`, and a throttled sequence from that stream `throttledStream`. I yield values into `stream` for a while, and eventually call `finish()` on its continuation. All the while in the background I am iterating through `throttledStream` - when `finish()` gets called on `stream`, my iteration through `throttledStream` finishes, but I don't get the final value that was sent to `stream`.

Sorry this is kind of a complicated way to explain the issue, but I will make a reproducible example project if this behavior sounds like it is indeed not intended!

gregcotten:
I should add, I guess it's the "penultimate" value I'm looking for - I realize that "nil" is the final value in a "finished" AsyncStream

twittemb:
Is it a throttle with “latest” ? If not then it can be normal that the penultimate element is skipped.

gregcotten:
> Is it a throttle with “latest” ? If not then it can be normal that the penultimate element is skipped.

Yes, latest = true, just calling `let throttledStream = stream.throttle(for: .seconds(1), latest: true)`, for example.

twittemb:
@gregcotten 

Keep in mind that AsyncSequence is a pull mechanism whereas Combine is a push mechanism for instance. It means that the throttling is applied to the pace of the consumer, not the producer.

in your case, let’s say you have a throttling duration of 1s, here is what I suspect is happening at the end of the timeline:

```
|
| x -> last received element by the consumer, a new throttling start time is set, ending in 1s
| y -> new element is discarded because it is too close to ‘x’ (the time elapsed since the start is < 1s)
|
| z -> last element is discarded because it is too close to ‘x’ (the time elapsed since the start is < 1s)
|
| nil -> the upstream sequence is finished before the theoretical end of this throttling session
|
| ——> theoretical end for the current throttling session (start + 1s)
V
```

In the end the consumer doesn’t receive the penultimate element but the last that matches un throttling session’s end, that is to say ‘x’.

I hope this helps. @FranzBusch could also chime in the confirm or not.

gregcotten:
Yes, this behavior makes sense to me, but I think it might be nice to find a way to always get the penultimate value somehow... Would a PR be welcome for this?

twittemb:
[...]

PR #296, opened 2023-09-21
Make throttle `internal`

FranzBusch:
# Motivation
During the discussion in https://github.com/apple/swift-async-algorithms/issues/248 it became clear that the semantics of `throttle` are not 100% figured out yet.

# Modification
This PR is making `throttle` internal to remove it from the upcoming 1.0.0 release. This gives us more time to investigate what exact semantics we want to have.


phausler:
I think this should probably be public but underscored.

FranzBusch:
@swift-ci please test

phausler:
@swift-ci please test

PR #298, opened 2023-09-24
Fix 5.7/5.8 build errors

FranzBusch:
# Motivation

Currently, this repo fails to build on Swift 5.7 and 5.8 since we were using clock APIs that were only available on 5.9.


FranzBusch:
@swift-ci please test

PR #303, opened 2023-11-02
Fix potential deadlocks when resuming a continuation while holding a lock

FranzBusch:
We wrongly assumed that it is safe to resume continuations while holding our state machine lock. There is a potential for a deadlock in rare circumstances. The problem is that the runtime holds a task specific lock. This lock is both acquired by task cancellation and continuation resumption. This can lead to scenarios where we are resuming the task's continuation while holding the state machine lock and a different thread cancelling the task and holding the runtime lock. Now both locks are held by different threads and they are waiting to acquire the opposite lock; hence, deadlocking the process.

In this PR, I went through all of our state machine based implementations and made sure that we are only acquiring the lock for as little as possible and handle all of the actions outside of the lock.

### DIFF
--- a/Sources/AsyncAlgorithms/Buffer/BoundedBufferStorage.swift
+++ b/Sources/AsyncAlgorithms/Buffer/BoundedBufferStorage.swift
@@ -18,34 +18,47 @@ final class BoundedBufferStorage<Base: AsyncSequence>: Sendable where Base: Send
 
   func next() async -> Result<Base.Element, Error>? {
     return await withTaskCancellationHandler {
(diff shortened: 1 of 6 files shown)


ole:
Out of curiosity, is this about the issue described in [Swift #67110: waitForStatusRecordUnlockIfNotSelfLocked deadlocks on macOS 14.0](https://github.com/apple/swift/issues/67110)?

FranzBusch:
@ole Yes this would fix the issue.

FranzBusch:
@swift-ci please test

ole:
@FranzBusch Thanks!

PR #305, opened 2023-12-19
Add `AsyncBackpressuredStream` proposal and implementation

FranzBusch:
# Motivation

The pitch to add external backpressure support to the standard libraries `AsyncStream` got returned for revision since there are larger open questions around `AsyncSequence`. However, having external backpressure in a source asynchronous sequence is becoming more and more important.

# Modification

This PR adds a modified proposal and implementation that brings the Swift Evolution proposal over to Swift Async Algorithms.


FranzBusch:
@swift-ci please test

Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerChannel+Internal.swift
```diff
@@ -0,0 +1,1409 @@
...
+                    switch action {
+                    case .resumeConsumerWithElement(let continuation, let element):
+                        continuation.resume(returning: element)
+
+                    case .resumeConsumerWithElementAndProducers(let continuation, let element, let producerContinuations):
+                        continuation.resume(returning: element)
+                        for producerContinuation in producerContinuations {
+                            switch producerContinuation {
+                            case .closure(let onProduceMore):
+                                onProduceMore(.failure(CancellationError()))
```
simonjbeaumont, line 431: This is a state machine bug I picked up running the OpenAPIURLSession tests with this type:

```suggestion
                                onProduceMore(.success(()))
```

Might be worth seeing if we can have similar tests in this project. This particular test, `1kChunk_10kChunks_100BDownloadWatermark`, exercised the suspend and resume of the producer a lot due to the relation of the chunk size to the watermark.

Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerChannel.swift
```diff
@@ -0,0 +1,489 @@
...
+    ///   - elementType: The element type of the channel.
+    ///   - failureType: The failure type of the channel.
+    ///   - BackpressureStrategy: The backpressure strategy that the channel should use.
+    /// - Returns: A tuple containing the channel and its source. The source should be passed to the
+    ///   producer while the channel should be passed to the consumer.
+    public static func makeChannel(
+        of elementType: Element.Type = Element.self,
+        throwing failureType: Failure.Type = Never.self,
+        backpressureStrategy: Source.BackpressureStrategy
+    ) -> ChannelAndStream {
```
simonjbeaumont, line 115: What's the reason for this returning a nominal type? Previously it returned a tuple which allows for initialization of the sequence as follows:

```swift
let (channel, source) = MultiProducerSingleConsumerChannel.makeChannel(/* ... */)
```

This is also what the documentation comment on the type suggests.

However, with the nominal type you cannot do that and, if you are storing these as properties, are forced to do this...

```swift
let channelAndSource = MultiProducerSingleConsumerChannel.makeChannel(/* ... */)
self.channel = channelAndSource.channel
self.source = channelAndSource.source
```

...instead of this...

```swift
(self.channel, self.source) = MultiProducerSingleConsumerChannel.makeChannel(/* ... */)
```

FranzBusch: Sadly, tuples don't support `~Copyable` types. That's the only reason for a nominal type here.

Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerChannel.swift
```diff
@@ -0,0 +1,489 @@
...
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=6.0)
```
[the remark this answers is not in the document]
FranzBusch: Hm not sure yet. Especially since it is part of the public API in the `Source`. Making the `Source` copyable makes the type behave way differently.

Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerChannel.swift
```diff
@@ -0,0 +1,489 @@
...
+        deinit {
+            storage.sequenceDeinitialized()
+        }
+    }
+
+    /// The backing storage.
+    private let backing: _Backing
+
+    @frozen
+    public struct ChannelAndStream: ~Copyable {
```
ser-0xff, line 90: Was it intended to call it `ChannelAndStream` while it is actually a container for `channel` and `source`?
Shouldn't it be called `ChannelAndSource`?

ser-0xff:
Thank you for the reply...
Unfortunately `internal import` does not work if we want to use package type (like `MultiProducerSingleConsumerChannel`) in the framework public API...

hassila:
> Thank you for the reply... Unfortunately `internal import` does not work if we want to use package type (like `MultiProducerSingleConsumerChannel`) in the framework public API...

@ser-0xff  I guess we need to wrap up the API surface we want to support then and just do internal import as suggested? But let's take that offline from this PR - thanks @FranzBusch for quick reply.

ser-0xff:
Hi, @FranzBusch 
We run into one another issue when we played with the `MultiProducerSingleConsumerChannel`.
Our projects are already on Swift6 which has a stricter non-copyable types checks. The  `MultiProducerSingleConsumerChannel<>.Source` type is non-copyable so we can't share it across many tasks, as a result we can use it only in `single producer` mode. How could we solve that issue and produce data within multiple tasks?

FranzBusch:
> We run into another issue when we played with the `MultiProducerSingleConsumerChannel`.
> Our projects are already on Swift6 which has a stricter non-copyable types checks. The `MultiProducerSingleConsumerChannel<>.Source` type is non-copyable so we can't share it across many tasks, as a result we can use it only in `single producer` mode.
> How could we solve that issue and produce data within multiple tasks?

The source has a `copy` method that allows you to get a second source. Now the problem is that closures cannot be marked as `~Copyable` or rathe `@calledOnce`. To solution to this is to box the source into a class right now and move it into a task. This either requires a lock or an `@unchecked Sendable` annotation right now.

ser-0xff:
Missed that `copy()` method in source code.
Will look at that, thank you for the reply.

ser-0xff:
HI, Franz!
We tried using `MultiProducersSingleConsumerChannel` instead of `AsyncStream` and encountered behavior unexpected for us. We expected the channel to remain functional as long as at least one associated source was active. However, in the current implementation, it appears that the channel transitions to a finished state as soon as any single source instance is deinitialized.
I created a [PR](https://github.com/FranzBusch/swift-async-algorithms/pull/2) with minimized test into your feature branch for a case if you could find some time to look at that and may be later will include that test to the test suite for `MultiProducersSingleConsumerChannel`.

FranzBusch:
> HI, Franz! We tried using `MultiProducersSingleConsumerChannel` instead of `AsyncStream` and encountered behavior unexpected for us. We expected the channel to remain functional as long as at least one associated source was active. However, in the current implementation, it appears that the channel transitions to a finished state as soon as any single source instance is deinitialized. I created a [PR](https://github.com/FranzBusch/swift-async-algorithms/pull/2) with minimized test into your feature branch for a case if you could find some time to look at that and may be later will include that test to the test suite for `MultiProducersSingleConsumerChannel`.

You are right. I haven't gotten around to implement the correct handling of multiple sources yet.

Sources/AsyncAlgorithms/Internal/_TinyArray.swift
```diff
@@ -0,0 +1,329 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// ``_TinyArray`` is a ``RandomAccessCollection`` optimised to store zero or one ``Element``.
+/// It supports arbitrary many elements but if only up to one ``Element`` is stored it does **not** allocate separate storage on the heap
+/// and instead stores the ``Element`` inline.
+@usableFromInline
+struct _TinyArray<Element> {
```
phausler, line 29: I wonder if this could perhaps be subsumed by `InlineArray` instead since the original intent was to have a faster storage (this is not a blocking concept)

Sources/AsyncAlgorithms/Locking.swift
```diff
@@ -17,24 +17,35 @@ import Glibc
 import WinSDK
 #endif
 
-internal struct Lock {
+@usableFromInline
+internal class Lock {
```
phausler, line 21: umm why is this a non-final class? If we are moving to a class then we should just use `Mutex` from Synchronization

Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerChannel.swift
```diff
@@ -0,0 +1,489 @@
...
+                    internalBackpressureStrategy: .unbounded(.init())
+                )
+            }
+        }
+
+        /// A type that indicates the result of sending elements to the source.
+        public enum SendResult: ~Copyable, Sendable {
+            /// A token that is returned when the channel's backpressure strategy indicated that production should
+            /// be suspended. Use this token to enqueue a callback by  calling the ``enqueueCallback(_:)`` method.
+            public struct CallbackToken: Sendable {
```
phausler, line 190: often tokens need to be hashable or at the bare min equatable; that might be worthwhile t consider.

Tests/AsyncAlgorithmsTests/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerChannelTests.swift
```diff
@@ -0,0 +1,1041 @@
...
+//===----------------------------------------------------------------------===//
+
+import AsyncAlgorithms
+import XCTest
+
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+final class MultiProducerSingleConsumerChannelTests: XCTestCase {
+    // MARK: - sequenceDeinitialized
+
+    // Following tests are disabled since the channel is not getting deinited due to a known bug
```
phausler, line 19: This might be a requirement to landing; id rather not introduce a type with a known leak

Evolution/0016-mutli-producer-single-consumer-channel.md
```diff
@@ -0,0 +1,652 @@
...
+the greatest performance. The callback only has to be allocated in the case
+where the producer needs to be suspended.
+
+Additionally, the above API is the building block for some higher-level and
+easier-to-use APIs to send values to the channel. Below is an
+example of the two higher-level APIs.
+
+```swift
+// Writing new values and providing a callback when to produce more
+try source.send(contentsOf: sequence, onProduceMore: { result in
```
phausler, line 184: can this be written without ambiguity as a trailing closure?

phausler (commented):
I think with some minor edits the pitch is good to go; id like to at the bare minimum see some of that testing reenabled to ensure it isn't leaking.

Sources/AsyncAlgorithms/Internal/_TinyArray.swift
FranzBusch: `InlineArray` isn't capable of fully replacing the need here. The idea behind `TinyArray` is to have a fast path for a single element that doesn't allocate and one that allocates if there are more. Since we can't tell at compile time how many producers we have we need this runtime dynamism.

Sources/AsyncAlgorithms/Locking.swift
FranzBusch: I removed all of the changes here since the latest code now uses `Mutex`.

Tests/AsyncAlgorithmsTests/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerChannelTests.swift
FranzBusch: I added significantly more tests now and also re-enabled those.

Evolution/0016-mutli-producer-single-consumer-channel.md
FranzBusch: Yes in a synchronous context both `send` methods can be called without ambiguity, e.g. this compiles:
```
try source.send(1)
source.send(2) { result in
    print(result)
}
```

The only thing that needs some special handling is if you want to call a sync send from an async context since the async methods are preferred; however, users can workaround this by creating an inline closure, e.g.:

```
try { try source.send(1) }()
```

Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerChannel.swift
```diff
@@ -0,0 +1,624 @@
...
+    public func setOnTerminationCallback(_ callback: @escaping @Sendable () -> Void) {
+      self._storage.onTermination = callback
+    }
+
+    /// Creates a new source which can be used to send elements to the channel concurrently.
+    ///
+    /// The channel will only automatically be finished if all existing sources have been deinited.
+    ///
+    /// - Returns: A new source for sending elements to the channel.
+    public mutating func copy() -> sending Self {
```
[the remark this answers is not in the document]
FranzBusch: Thanks for calling this out. I think the right thing is to actually remove the `Sendable` conformance on `Source` since it is in fact not `Sendable`.

Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerChannel.swift
```diff
@@ -0,0 +1,624 @@
...
+    /// A struct representing the backpressure of the channel.
+    public struct BackpressureStrategy: Sendable {
+      var internalBackpressureStrategy: _InternalBackpressureStrategy
+
+      /// A backpressure strategy using a high and low watermark to suspend and resume production respectively.
+      ///
+      /// - Parameters:
+      ///   - low: When the number of buffered elements drops below the low watermark, producers will be resumed.
+      ///   - high: When the number of buffered elements rises above the high watermark, producers will be suspended.
+      public static func watermark(low: Int, high: Int) -> BackpressureStrategy {
```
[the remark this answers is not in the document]
FranzBusch: I think it's interesting but I personally prefer having to `Int`'s here since the watermark strategy is not really about the range but a low and high limit where the production is started and stopped respectively.

Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerChannel+Internal.swift
```diff
@@ -0,0 +1,1666 @@
...
+        self._low = low
+        self._high = high
+        self._waterLevelForElement = waterLevelForElement
+      }
+
+      @inlinable
+      mutating func didSend(elements: Deque<Element>.SubSequence) -> Bool {
+        if let waterLevelForElement = self._waterLevelForElement {
+          for element in elements {
+            self._currentWatermark += waterLevelForElement(element)
```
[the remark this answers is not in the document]
FranzBusch: I updated the public docs to call this out. I think the reality is that we need to call this code while being under a lock since the state management otherwise would become almost impossible with potentially any code interleaving.

Evolution/0016-mutli-producer-single-consumer-channel.md
```diff
@@ -0,0 +1,847 @@
...
+## Detailed design
+
+```swift
+#if compiler(>=6.1)
+/// An error that is thrown from the various `send` methods of the
+/// ``MultiProducerSingleConsumerAsyncChannel/Source``.
+///
+/// This error is thrown when the channel is already finished when
+/// trying to send new elements to the source.
+public struct MultiProducerSingleConsumerAsyncChannelAlreadyFinishedError: Error { }
```
phausler, line 401: This seems like it should be a nested type; it is a keyboard full of naming here...

Evolution/0016-mutli-producer-single-consumer-channel.md
```diff
@@ -0,0 +1,847 @@
...
+leads to the termination of all consumers. This is not expected from multicast
+asynchronous sequences in general.
+
+## Proposed solution
+
+The above motivation lays out the expected behaviors for any consumer/producer
+system and compares them to the behaviors of `Async[Throwing]Stream` and
+`Async[Throwing]Channel`.
+
+This section proposes a new type called `MultiProducerSingleConsumerAsyncChannel`
```
phausler, line 132: Id really love to come up with a smaller name. We have been using water metaphors, so perhaps something in that family of names? From my understanding of the pitch it is like you have many small tributaries contributing to a larger stream of values. Id like to come up with some alternatives than this.

Evolution/0016-mutli-producer-single-consumer-channel.md
```diff
@@ -0,0 +1,847 @@
...
+        }
+
+        /// A type that indicates the result of sending elements to the source.
+        public enum SendResult: ~Copyable, Sendable {
+            /// An opaque token that is returned when the channel's backpressure strategy indicated that production should
+            /// be suspended. Use this token to enqueue a callback by  calling the ``MultiProducerSingleConsumerAsyncChannel/Source/enqueueCallback(callbackToken:onProduceMore:)`` method.
+            ///
+            /// - Important: This token must only be passed once to ``MultiProducerSingleConsumerAsyncChannel/Source/enqueueCallback(callbackToken:onProduceMore:)``
+            ///  and ``MultiProducerSingleConsumerAsyncChannel/Source/cancelCallback(callbackToken:)``.
+            public struct CallbackToken: Sendable, Hashable { }
```
phausler: Is it possible that these tokens are somehow user provided?

Evolution/0016-mutli-producer-single-consumer-channel.md
```diff
@@ -0,0 +1,847 @@
...
+            /// Indicates that a callback should be enqueued.
+            ///
+            /// The associated token should be passed to the ````MultiProducerSingleConsumerAsyncChannel/Source/enqueueCallback(callbackToken:onProduceMore:)```` method.
+            case enqueueCallback(CallbackToken)
+        }
+
+        /// A callback to invoke when the channel finished.
+        ///
+        /// This is called after the last element has been consumed by the channel.
+        public func setOnTerminationCallback(_ callback: @escaping @Sendable () -> Void)
```
phausler, line 621: it might be useful to have a parameter in the closure to tell why the termination happened, either finished, failed, or cancelled.

Evolution/0016-mutli-producer-single-consumer-channel.md
FranzBusch: I agree it is a long but we can't nest it due to the `MPSCCAsyncChannel` being generic.

Evolution/0016-mutli-producer-single-consumer-channel.md
FranzBusch: What if we call it `MPSCAsyncChannel`. The German in me does like long explicit wordings but I'm happy to go with abbreviations here. The other name I was thinking about was `MultiProducerAsyncChannel` or `ExternalBackpressuredAsyncChannel`.

Evolution/0016-mutli-producer-single-consumer-channel.md
FranzBusch: If you mean constructed by the user then no. There is no public `init` for those.

Evolution/0016-mutli-producer-single-consumer-channel.md
pyrtsa: `MultiProducerAsyncChannel` sounds good to me!

Evolution/0016-mutli-producer-single-consumer-channel.md
phausler: no, I was suggesting that the user provides a type to represent a token and this could be token that the framework provides. It would mean that things would need to be flipped around to be generic around that.

There are some APIs that have tokens associated with callbacks. Thinking particularly the CF APIs or the KVO APIs

Evolution/0016-mutli-producer-single-consumer-channel.md
phausler: 4 words is better than 6 I guess...

Evolution/0016-mutli-producer-single-consumer-channel.md
phausler: I guess folks aren't going to be typing this type out much eh, so perhaps not a huge deal; do we really even need this type to be public?

Evolution/0016-mutli-producer-single-consumer-channel.md
FranzBusch: I see what you mean I would like to avoid that though since we are using those tokens to understand if there is a potential API mis-use. We wouldn't be able to do this if the user were to create those tokens.

Evolution/0016-mutli-producer-single-consumer-channel.md
```diff
@@ -0,0 +1,838 @@
...
+regulating the flow of data between different components. Additionally, it
+allows to put an upper bound on resource consumption of a system. In reality,
+backpressure is used in almost all networked applications.
+
+In Swift, asynchronous sequence also have the concept of internal backpressure.
+This modeled by the pull-based implementation where a consumer has to call
+`next` on the `AsyncIterator`. In this model, there is no way for a consumer to
+overwhelm a producer since the producer controls the rate of pulling elements.
+
+However, the internal backpressure of an asynchronous isn't the only
```
heckj:
```suggestion
However, the internal backpressure of an asynchronous sequence isn't the only
```

"asynchronous" what - seems to be missing a word, I think it's meant to be "stream" or "sequence" - and I'd guess sequence from the context

Evolution/0016-mutli-producer-single-consumer-channel.md
```diff
@@ -0,0 +1,838 @@
...
+
+## Comparison to other root asynchronous primitives
+
+### swift-async-algorithm: AsyncChannel
+
+The `AsyncChannel` is a multi-consumer/multi-producer root asynchronous sequence
+which can be used to communicate between two tasks. It only offers asynchronous
+production APIs and has an effective buffer of one per producer. This means that
+any producer will be suspended until its value has been consumed. `AsyncChannel`
+can handle multiple consumers and resumes them in FIFO order.
```
alessionossa, line 752: There is some uncertainty about the FIFO ordering of `AsyncChannel`. See https://github.com/apple/swift-async-algorithms/issues/356

PR #307, opened 2024-02-09
Depend on specific products from swift-collections

erichoracek:
Before this change, once consumers update to SwiftCollections 1.1.0 they'll pull in all of the new collection dependencies (HashTree, BitCollections, etc.). In our case, this increased our binary size by ~1MB. To fix this, we switch to only depending on what we need.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -20,7 +20,7 @@ let package = Package(
   targets: [
     .target(
       name: "AsyncAlgorithms",
-      dependencies: [.product(name: "Collections", package: "swift-collections")],
+      dependencies: [.product(name: "OrderedCollections", package: "swift-collections")],
       swiftSettings: [
           .enableExperimentalFeature("StrictConcurrency=complete"),
       ]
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -20,7 +20,7 @@ let package = Package(
   targets: [
     .target(
       name: "AsyncAlgorithms",
-      dependencies: [.product(name: "Collections", package: "swift-collections")],
+      dependencies: [.product(name: "OrderedCollections", package: "swift-collections")],
```
FranzBusch, line 23: I don't think this is enough. We also need to depend on `DequeModule` since we use `Deque` in quite a few places

FranzBusch (changes requested):

Package.swift
erichoracek: Ah I didn't realize, added in d685a5086ccb9decfa6dc83a8558c5c5911559b0

FranzBusch:
@swift-ci please test

FranzBusch (approved):

PR #310, opened 2024-04-04
Fix some strict concurrency warnings

FranzBusch:
# Motivation

There were a few new strict concurrency warnings that this PR fixes.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,7 +14,7 @@ let package = Package(
     .library(name: "AsyncAlgorithms", targets: ["AsyncAlgorithms"]),
   ],
   dependencies: [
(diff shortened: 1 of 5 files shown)


FranzBusch:
@swift-ci please test

phausler (approved):
I presume if there is then later a `transfer` keyword we will likely migrate to that eh?

FranzBusch:
> I presume if there is then later a `transfer` keyword we will likely migrate to that eh?

Sadly we can't since the elements are returned from an iterator. The next method is not marked as transferring the return element so we can't forward transfer it. However, what we are doing here is safe since the elements are never owned by more than one task.

PR #311, opened 2024-04-10
Add support for `SWIFTCI_USE_LOCAL_DEPS` convention

kateinoigakukun:
This package is used by https://github.com/apple/swift-sdk-generator and it needs to be built during build-script pipeline, so we need to make its dependencies follow the CI convention. CC: @MaxDesiatov

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -13,10 +13,6 @@ let package = Package(
   products: [
     .library(name: "AsyncAlgorithms", targets: ["AsyncAlgorithms"]),
   ],
-  dependencies: [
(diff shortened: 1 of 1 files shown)


kateinoigakukun:
@swift-ci test

phausler (approved):

MaxDesiatov (approved):
Thanks!

kateinoigakukun:
@phausler Would you mind having a new tag to pin the version including this fix in update-checkout-config.json if possible? :pray:

Issue #247, opened 2023-01-27
Provide an option for chunking based on a signal to include empty chunks


tachyonics:
For sequences that chunk based on a signal, the signal will be effectively ignored if no elements were received since the previous signal. Adding an option to pass an empty chunk on such a signal would be useful, particularly for use cases where you want to detect and report on the absence of something regularly happening. 

An alternative approach would be to make this the default behaviour since there are other mechanisms that can be used to compose a sequence that only produces non-empty chunks.

EngOmarElsayed:
I could like to work on this issue, Can you provide more context on how to start and which part you want me to enhance ? @FranzBusch

PR #312, opened 2024-04-19
async chunked sequences can produce empty chunks

rnro:
async chunked sequences can produce empty chunks if signals are received before primary chunk elements.

This commit also includes an accompanying evolution proposal.


rnro:
This should resolve https://github.com/apple/swift-async-algorithms/issues/247

rnro:
Forums post on this https://forums.swift.org/t/swift-async-algorithms-proposal-produce-empty-chunks/72575

PR #317, opened 2024-06-18
Fix compile errors for Swift 6 language mode

b1ackturtle:
This PR fixes code that caused compile errors when enabling the Swift 6 language mode.


Sources/AsyncSequenceValidation/TaskDriver.swift
```diff
@@ -20,12 +20,12 @@ import Glibc
 #endif
 
 #if canImport(Darwin)
-func start_thread(_ raw: UnsafeMutableRawPointer) -> UnsafeMutableRawPointer? {
+@Sendable func start_thread(_ raw: UnsafeMutableRawPointer) -> UnsafeMutableRawPointer? {
```
FranzBusch, line 23: Adding `@Sendable` to functions isn't the right fix often. Where did the original error come from?

Sources/AsyncSequenceValidation/Test.swift
```diff
@@ -412,3 +412,32 @@ extension AsyncSequenceValidationDiagram {
     try self.test(theme: .ascii, build)
   }
 }
+
+final class LockIsolated<Value>: @unchecked Sendable {
```
FranzBusch, line 416: We do have this type already called `ManagedCriticalState`

Sources/AsyncSequenceValidation/Test.swift
```diff
@@ -116,19 +116,19 @@ extension AsyncSequenceValidationDiagram {
...
+      _executor.asUnownedSerialExecutor()
     }
 #endif
 
-    static var clock: Clock?
-    
-    
-    
-    static var driver: TaskDriver?
+    static let clock = LockIsolated<Clock?>(nil)
```
FranzBusch, line 123: Having multiple separate locks here to protect the state is problematic since there might be invariants that aren't enforced this way. I would rather put all the state behind one lock

Sources/AsyncSequenceValidation/TaskDriver.swift
b1ackturtle: This `@Sendable` is intended to address the compilation error that occurs in `TaskDriver`'s `pthread_create(&thread, nil, start_thread, Unmanaged.passRetained(self).toOpaque())` when the Swift 6 language mode is enabled.
The compilation error that occurs is as follows:

> A C function pointer can only be formed from a reference to a 'func' or a literal closure

After reconsidering based on your comments, I realized that defining `start_thread` within `start` can also suppress the compilation error.
I have tried fixing it with [ea58d5d](https://github.com/apple/swift-async-algorithms/pull/317/commits/ea58d5da0942db511abc9e8b424faec9766cc60d).

Sources/AsyncSequenceValidation/Test.swift
b1ackturtle: I did not realize that `ManagedCriticalState` was already defined. I have fixed it to use `ManagedCriticalState`.

Sources/AsyncSequenceValidation/Test.swift
b1ackturtle: I have modified it to protect all states with a single lock.

FranzBusch:
@swift-ci please test

PR #320, opened 2024-06-27
Bump swift-collections to 1.1.1

yutailang0119:
## Information

- Package version: 1.0.1
- Swift version: 5.10

### Behavior

The app upload to App Store Connect fails due to swift-collections, when not using `SWIFTCI_USE_LOCAL_DEPS`.
This issue has been reported in https://github.com/apple/swift-collections/issues/364.
This package needs to be updated to version 1.1.1.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -51,7 +51,7 @@ let package = Package(
 
 if Context.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
   package.dependencies += [
-    .package(url: "https://github.com/apple/swift-collections.git", from: "1.1.0"),
+    .package(url: "https://github.com/apple/swift-collections.git", from: "1.1.1"),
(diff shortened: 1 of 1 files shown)


FranzBusch:
It shouldn't be necessary for us to increase the minimum version here. In your application you should be able to update to the latest package versions which should pull in the `1.1.1` release if `swift-collections`.

yutailang0119:
In many cases, that is correct, but it can cause issues due to caches and other factors.
Since I know there are problems with swift-collections v1.1.0, it would be helpful to explicitly update the minimum version.

FranzBusch:
I understand how it can still result in build failures but I disagree on bumping the version here. This isn't going to be sustainable across the ecosystem that every single package is bumping their minimum version for a build fix from one of its dependencies.

yutailang0119:
OK, I will consider explicitly adding a dependency on swift-collections v1.1.1 in my app.
Thanks!

PR #324, opened 2024-07-30
Implement AsyncMapErrorSequence

clive819:
Pretty much the same as [Combine.Publisher.mapError](https://developer.apple.com/documentation/combine/publisher/maperror(_:)) but for AsyncSequence.

This enables you to do something like this:
```swift
let sequence = myAsyncSequence.mapError { MyGenericError(wrappedError: $0) }
```

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
@@ -1,0 +1,78 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+public extension AsyncSequence {
+
+    /// Converts any failure into a new error.
+    ///
+    /// - Parameter transform: A closure that takes the failure as a parameter and returns a new error.
+    /// - Returns: An asynchronous sequence that maps the error thrown into the one produced by the transform closure.
+    ///
+    /// Use the ``mapError(_:)`` operator when you need to replace one error type with another.
+    func mapError<ErrorType>(transform: @Sendable @escaping (Error) -> ErrorType) -> AsyncMapErrorSequence<Self, ErrorType> {
+        .init(base: self, transform: transform)
+    }
+}
+
+/// An asynchronous sequence that converts any failure into a new error.
+public struct AsyncMapErrorSequence<Base: AsyncSequence, ErrorType: Error>: AsyncSequence, Sendable where Base: Sendable {
+
+    public typealias AsyncIterator = Iterator
+    public typealias Element = Base.Element
+
+    private let base: Base
+    private let transform: @Sendable (Error) -> ErrorType
+
+    init(
+        base: Base,
+        transform: @Sendable @escaping (Error) -> ErrorType
+    ) {
+        self.base = base
+        self.transform = transform
+    }
+
+    public func makeAsyncIterator() -> Iterator {
+        Iterator(
+            base: base.makeAsyncIterator(),
+            transform: transform
+        )
+    }
+}
+
+public extension AsyncMapErrorSequence {
+    
+    /// The iterator that produces elements of the map sequence.
+    struct Iterator: AsyncIteratorProtocol {
+
+        public typealias Element = Base.Element
+
+        private var base: Base.AsyncIterator
+
+        private let transform: @Sendable (Error) -> ErrorType
+
+        public init(
+            base: Base.AsyncIterator,
+            transform: @Sendable @escaping (Error) -> ErrorType
+        ) {
+            self.base = base
+            self.transform = transform
+        }
+
+        public mutating func next() async throws -> Element? {
+            do {
+                return try await base.next()
+            } catch {
+                throw transform(error)
+            }
+        }
+    }
+}
--- a/Tests/AsyncAlgorithmsTests/TestMapError.swift
+++ b/Tests/AsyncAlgorithmsTests/TestMapError.swift
@@ -1,0 +1,83 @@
+import AsyncAlgorithms
+import XCTest
+
+final class TestMapError: XCTestCase {
+
+    func test_mapError() async throws {
+        let array = [URLError(.badURL)]
+        let sequence = array.async
+            .map { throw $0 }
+            .mapError { _ in
+                MyAwesomeError()
+            }
+
+        do {
+            for try await _ in sequence {
+                XCTFail("sequence should throw")
+            }
+        } catch {
+            _ = try XCTUnwrap(error as? MyAwesomeError)
+        }
+    }
+
+    func test_nonThrowing() async throws {
+        let array = [1, 2, 3, 4, 5]
+        let sequence = array.async
+            .mapError { _ in
+                MyAwesomeError()
+            }
+
+        var actual: [Int] = []
+        for try await value in sequence {
+            actual.append(value)
+        }
+        XCTAssertEqual(array, actual)
+    }
+
+    func test_cancellation() async throws {
+        let source = Indefinite(value: "test").async
+        let sequence = source.mapError { _ in MyAwesomeError() }
+
+        let finished = expectation(description: "finished")
+        let iterated = expectation(description: "iterated")
+
+        let task = Task {
+            var firstIteration = false
+            for try await el in sequence {
+                XCTAssertEqual(el, "test")
+
+                if !firstIteration {
+                    firstIteration = true
+                    iterated.fulfill()
+                }
+            }
+            finished.fulfill()
+        }
+
+        // ensure the other task actually starts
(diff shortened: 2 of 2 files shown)


Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
```diff
@@ -0,0 +1,78 @@
...
+    /// - Returns: An asynchronous sequence that maps the error thrown into the one produced by the transform closure.
+    ///
+    /// Use the ``mapError(_:)`` operator when you need to replace one error type with another.
+    func mapError<ErrorType>(transform: @Sendable @escaping (Error) -> ErrorType) -> AsyncMapErrorSequence<Self, ErrorType> {
+        .init(base: self, transform: transform)
+    }
+}
+
+/// An asynchronous sequence that converts any failure into a new error.
+public struct AsyncMapErrorSequence<Base: AsyncSequence, ErrorType: Error>: AsyncSequence, Sendable where Base: Sendable {
```
phausler, line 27: The `Sendable` part should only be conditional when the `Base` is `Sendable` (not a constraint)

Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    init(
+        base: Base,
+        transform: @Sendable @escaping (Error) -> ErrorType
+    ) {
+        self.base = base
+        self.transform = transform
+    }
+
+    public func makeAsyncIterator() -> Iterator {
```
phausler, line 43: This is returning an internal type from a public function.

Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
```diff
@@ -0,0 +1,78 @@
...
+
+public extension AsyncSequence {
+
+    /// Converts any failure into a new error.
+    ///
+    /// - Parameter transform: A closure that takes the failure as a parameter and returns a new error.
+    /// - Returns: An asynchronous sequence that maps the error thrown into the one produced by the transform closure.
+    ///
+    /// Use the ``mapError(_:)`` operator when you need to replace one error type with another.
+    func mapError<ErrorType>(transform: @Sendable @escaping (Error) -> ErrorType) -> AsyncMapErrorSequence<Self, ErrorType> {
```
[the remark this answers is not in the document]
FranzBusch: Should we return a `some AsyncSequence` instead of a concrete one here as well?

Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
```diff
@@ -0,0 +1,78 @@
...
+
+        public init(
+            base: Base.AsyncIterator,
+            transform: @Sendable @escaping (Error) -> ErrorType
+        ) {
+            self.base = base
+            self.transform = transform
+        }
+
+        public mutating func next() async throws -> Element? {
```
FranzBusch, line 70: Should we adopt typed throws and the `isolated` parameter here right away since this is a new type?

Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
clive819: `Iterator` is inside the public extension so it's public, but I can add `public` to it just to make it more explicit.

Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
clive819: Sure, but we'd need to bump the package Swift version to 6, do we want to do that?

FranzBusch: We don't have to. We should be able to just provide that in a compiler conditional block with `#if compiler(>6.0)`

Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
FranzBusch: In general we prefer to avoid access level on extensions and put it on each individual member to make it clearer what access level each method and type has

Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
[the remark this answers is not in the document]
FranzBusch: Yeah they predate `AsyncSequence` adopting primary associated types so we can do that now. It should also allow us to avoid making the actual types public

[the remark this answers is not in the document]
clive819: I decided not to return `some AsyncSequence` and return concrete type instead because using an opaque type would render typed throws ineffective by erasing the type, thereby preventing the compiler from ensuring that the returned sequence matches our intended new type. Consequently, it would simply default to any Error. And I believe the benefits of using typed throws for this specific case outweight the exposure of the internal types.

Additionally, I didn't use `Failure` since it's only available in iOS 18+, I'd need 2 functions to make it backwards compatible. And `AsyncSequence.Failure` is actuall `any Error` - there's no benefit of using `Failure` v.s. `any Error` since they are the same. (side note: `AsyncIteratorProtocol.next()` didn't adopt typed throws but `AsyncIteratorProtocol.next(isolation:)` does adopt typed throws. Looks like intentional to maintain backward compatibility. But it's also one of the reasons why I didn't use `Failure`.) You can checkout this commit to see my attempt https://github.com/clive819/swift-async-algorithms/commit/a8f139a49dc9efa7657993dd52782d98ed9c6d4b#diff-39026a27cb7e7c71fcd5df08416cb255085e46b3a227abecdc820d13a0307ab0R12-R15

Tests/AsyncAlgorithmsTests/TestMapError.swift
```diff
@@ -0,0 +1,89 @@
...
+
+        do {
+            for try await _ in sequence {
+                XCTFail("sequence should throw")
+            }
+        } catch {
+#if compiler(>=6.0)
+            // NO-OP
+            // The compiler already checks that for us since we're using typed throws.
+            // Writing that assert will just give compiler warning.
```
clive819, line 22: AFAIK, there's no way to expect compiler warnings, so I just make it a no-op block.

Sources/AsyncAlgorithms/AsyncMapErrorSequence.swift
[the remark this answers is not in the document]
clive819: Address in https://github.com/apple/swift-async-algorithms/pull/324/commits/4ba9c84becdc3279ced63bf2faea6ce705a9769d

Also created a post for it https://forums.swift.org/t/pitch-map-error/75422

Evolution/NNNN-map-failure.md
```diff
@@ -0,0 +1,149 @@
...
+* Bugs: 
+
+## Introduction
+
+The `mapFailure` function empowers developers to elegantly transform errors within asynchronous sequences, enhancing code readability and maintainability.
+
+```swift
+extension AsyncSequence {
+
+    public func mapFailure<MappedFailure: Error>(_ transform: @Sendable @escaping (Self.Failure) -> MappedFailure) -> some AsyncSequence<Self.Element, MappedFailure> {
```
FranzBusch, line 21: We need to make this probably two methods to account for `Sendable`.
```
public func mapFailure<MappedFailure: Error>(_ transform: @Sendable @escaping (Self.Failure) -> MappedFailure) -> some AsyncSequence<Self.Element, MappedFailure>


public func mapFailure<MappedFailure: Error>(_ transform: @Sendable @escaping (Self.Failure) -> MappedFailure) -> some (AsyncSequence<Self.Element, MappedFailure> & Sendable) where Self: Sendable
```

Evolution/NNNN-map-failure.md
```diff
@@ -0,0 +1,149 @@
...
+    ///
+    /// Use the ``mapError(_:)`` operator when you need to replace one error type with another.
+    @available(macOS, deprecated: 15.0, renamed: "mapFailure")
+    @available(iOS, deprecated: 18.0, renamed: "mapFailure")
+    @available(watchOS, deprecated: 11.0, renamed: "mapFailure")
+    @available(tvOS, deprecated: 18.0, renamed: "mapFailure")
+    @available(visionOS, deprecated: 2.0, renamed: "mapFailure")
+    public func mapError<MappedError: Error>(_ transform: @Sendable @escaping (any Error) -> MappedError) -> AsyncMapErrorSequence<Self, MappedError> {
+        .init(base: self, transform: transform)
+    }
```
FranzBusch, line 61: I don't think we need this. It is okay if we just ship this new type when the compiler is `6.0` and above

FranzBusch (changes requested):

Sources/AsyncAlgorithms/AsyncMapFailureSequence.swift
```diff
@@ -0,0 +1,109 @@
...
+
+        init(
+            base: Base.AsyncIterator,
+            transform: @Sendable @escaping (Failure) -> MappedFailure
+        ) {
+            self.base = base
+            self.transform = transform
+        }
+
+        mutating func next() async throws(MappedFailure) -> Element? {
```
phausler, line 86: very minor nit: you could just have this `next` function call the `next(isolation: nil)` to save on potential refactoring points.

not a blocker imho

clive819: oh yes! updated!

PR #273, opened 2023-06-23
Add WebAssembly support

0xpablo:
Hi everyone,

This PR adds support for WebAssembly (and possibly other single threaded environments).
I made the primitive an Int instead of a Void to avoid the `UnsafeMutablePointer<Void> has been replaced by UnsafeMutableRawPointer` warning.

<details>
  <summary>Build and tests running on SwiftWasm on Linux</summary>

```
root@684f97f7b63f:/# swift --version
SwiftWasm Swift version 5.8 (swiftlang-5.8.0)
Target: x86_64-unknown-linux-gnu
root@684f97f7b63f:/# cd saa/
root@684f97f7b63f:/saa# swift build
Building for debugging...
[145/145] Merging module AsyncAlgorithms_XCTest
Build complete! (58.26s)
root@684f97f7b63f:/saa# swift test 
Building for debugging...
[50/50] Linking swift-async-algorithmsPackageTests.xctest
Build complete! (32.33s)
Test Suite 'All tests' started at 2023-06-23 16:47:58.676
Test Suite 'debug.xctest' started at 2023-06-23 16:47:58.759
Test Suite 'TestAdjacentPairs' started at 2023-06-23 16:47:58.771
Test Case 'TestAdjacentPairs.test_adjacentPairs_finishes_when_iteration_task_is_cancelled' started at 2023-06-23 16:47:58.776
Test Case 'TestAdjacentPairs.test_adjacentPairs_finishes_when_iteration_task_is_cancelled' passed (0.048 seconds)
Test Case 'TestAdjacentPairs.test_adjacentPairs_forwards_termination_from_source_when_iteration_is_finished' started at 2023-06-23 16:47:58.825
Test Case 'TestAdjacentPairs.test_adjacentPairs_forwards_termination_from_source_when_iteration_is_finished' passed (0.004 seconds)
Test Case 'TestAdjacentPairs.test_adjacentPairs_produces_empty_sequence_when_source_sequence_is_empty' started at 2023-06-23 16:47:58.829
Test Case 'TestAdjacentPairs.test_adjacentPairs_produces_empty_sequence_when_source_sequence_is_empty' passed (0.005 seconds)
Test Case 'TestAdjacentPairs.test_adjacentPairs_produces_tuples_of_adjacent_values_of_original_element' started at 2023-06-23 16:47:58.834
Test Case 'TestAdjacentPairs.test_adjacentPairs_produces_tuples_of_adjacent_values_of_original_element' passed (0.005 seconds)
[...]

PR #325, opened 2024-08-05
Add Musl import, error if unrecognised platform

adam-fowler:
Add support for Musl
Also error if platform is not recognised.

### DIFF
--- a/Sources/AsyncAlgorithms/Locking.swift
+++ b/Sources/AsyncAlgorithms/Locking.swift
@@ -13,14 +13,18 @@
 import Darwin
 #elseif canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
 #elseif canImport(WinSDK)
 import WinSDK
+#else
+#error("Unsupported platform")
 #endif
 
 internal struct Lock {
 #if canImport(Darwin)
   typealias Primitive = os_unfair_lock
-#elseif canImport(Glibc)
+#elseif canImport(Glibc) || canImport(Musl)
   typealias Primitive = pthread_mutex_t
 #elseif canImport(WinSDK)
   typealias Primitive = SRWLOCK
@@ -38,7 +42,7 @@ internal struct Lock {
   fileprivate static func initialize(_ platformLock: PlatformLock) {
 #if canImport(Darwin)
     platformLock.initialize(to: os_unfair_lock())
-#elseif canImport(Glibc)
+#elseif canImport(Glibc) || canImport(Musl)
     let result = pthread_mutex_init(platformLock, nil)
     precondition(result == 0, "pthread_mutex_init failed")
 #elseif canImport(WinSDK)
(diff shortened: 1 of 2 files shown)


Sources/AsyncAlgorithms/Locking.swift
```diff
@@ -13,14 +13,18 @@
...
 #endif
 
 internal struct Lock {
 #if canImport(Darwin)
   typealias Primitive = os_unfair_lock
-#elseif canImport(Glibc)
+#elseif canImport(Glibc) || canImport(Musl)
   typealias Primitive = pthread_mutex_t
 #elseif canImport(WinSDK)
   typealias Primitive = SRWLOCK
```
FranzBusch, line 30: Can we make the `#else` case fail with an `#error` to avoid running into this again

adam-fowler: I did add it at the top, but can add here as well

FranzBusch (approved):

adam-fowler:
@finagolfin you may be interested in this.

adam-fowler:
It was more about the package not compiling for Android anymore when it did before.

finagolfin:
> It was more about the package not compiling for Android anymore when it did before.

OK, this package should cross-compile fine for Android with the current 5.10 release, but the upcoming 6.0 release uses a new Android overlay, that this package has not been ported to yet.

I have the trivial patch needed for that available locally on my Android phone, just need to clean it up and submit it.

adam-fowler:
> OK, this package should cross-compile fine for Android with the current 5.10 release

Ah that was the problem I found with Musl. It compiled fine, but the actual code caused crashes given the lock primitive defaulted to `Int` if the platform was unrecognised. https://github.com/apple/swift-async-algorithms/blob/6ae9a051f76b81cc668305ceed5b0e0a7fd93d20/Sources/AsyncAlgorithms/Locking.swift#L28

FranzBusch:
@swift-ci please test

Sources/AsyncAlgorithms/Locking.swift
```diff
@@ -13,19 +13,23 @@
...
 #else
-  typealias Primitive = Int
+  #error("Unsupported platform")
```
dnadoba, line 32: From gitblame it seem like this was added for web assembly:
https://github.com/apple/swift-async-algorithms/pull/273/files
[Which seems that this can be checked for with ](https://github.com/swiftlang/swift/blob/34588acedb285cf02c6539db6fa1036d54237c6e/test/Parse/ConditionalCompilation/wasm32Target.swift#L4)arch(wasm32)[ or ](https://github.com/swiftlang/swift/blob/34588acedb285cf02c6539db6fa1036d54237c6e/test/Parse/ConditionalCompilation/wasm32Target.swift#L4)os(WASI)
[There is also an underscore check for multithreading](https://github.com/swiftlang/swift/blob/34588acedb285cf02c6539db6fa1036d54237c6e/test/Parse/ConditionalCompilation/wasm32ThreadsTarget.swift#L4): #if _runtime(_multithreaded)
This was added in https://github.com/swiftlang/swift/pull/72649

swift-testing uses a combination of os(WASI) and _runtime(_multithreaded) combined with compiler(>=6.1)
https://github.com/swiftlang/swift-testing/pull/322/files

FWIW Mutex has just landed in Swift 6.0

dnadoba: cc: @0xpablo

PR #326, opened 2024-08-06
Use Bionic module from new Android overlay in Swift 6 instead

finagolfin:
The new module and overlay were merged into Swift 6 in swiftlang/swift#74758.

I've been building and testing this package natively on Android for some time, but the addition of nullability annotations to Bionic with NDK 26 last year and the new Android overlay in the Swift compiler this summer required these changes.

### DIFF
--- a/Sources/AsyncSequenceValidation/TaskDriver.swift
+++ b/Sources/AsyncSequenceValidation/TaskDriver.swift
@@ -17,6 +17,8 @@ import Darwin
 import Glibc
 #elseif canImport(Musl)
 import Musl
+#elseif canImport(Bionic)
+import Bionic
 #elseif canImport(WinSDK)
 #error("TODO: Port TaskDriver threading to windows")
 #else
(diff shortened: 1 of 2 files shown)


Sources/AsyncSequenceValidation/TaskDriver.swift
```diff
@@ -28,11 +30,16 @@ func start_thread(_ raw: UnsafeMutableRawPointer) -> UnsafeMutableRawPointer? {
   Unmanaged<TaskDriver>.fromOpaque(raw).takeRetainedValue().run()
   return nil
 }
-#elseif canImport(Glibc) || canImport(Musl)
+#elseif (canImport(Glibc) && !os(Android)) || canImport(Musl)
```
finagolfin, line 33: These two changes get `start_thread()` working with Swift 5.10 and NDK 26 also, by changing the signature for the nullability annotations. That's why this is the only portion of this pull to check `os(Android)`, because that's available in both Swift 5.10 and 6.

FranzBusch (approved):

PR #345, opened 2025-03-28
Move platform requirements to availability annotations

glbrntt:
Adding or raising the deployment platforms in the package manifest is a SemVer major breaking change as consumers must also add or raise their deployment platforms. This is a known limitation of SwiftPM.

Unforunately this means that it's very difficult for non-leaf packages to adopt packages which declare their platforms in the manifest. Doing so puts the brakes on adoption and ecosystem growth. For 'core' packages like this one availability constraints should be expressed on declarations rather than in the manifest.

This patch adds equivalent availability annotations to declarations across the package and removes platforms from the package manifest.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,12 +4,6 @@ import PackageDescription
 
 let package = Package(
   name: "swift-async-algorithms",
-  platforms: [
-    .macOS("10.15"),
-    .iOS("13.0"),
-    .tvOS("13.0"),
-    .watchOS("6.0"),
-  ],
   products: [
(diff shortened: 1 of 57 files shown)


FranzBusch (approved):

FranzBusch:
Let's get the CI moved to GitHub Actions and then we can test the PR with that.

FranzBusch:
@glbrntt can you rebase this PR?

glbrntt:
Closing in favour of #348

PR #348, opened 2025-04-08
Move platform requirements to availability annotations

glbrntt:
Adding or raising the deployment platforms in the package manifest is a SemVer major breaking change as consumers must also add or raise their deployment platforms. This is a known limitation of SwiftPM.

Unforunately this means that it's very difficult for non-leaf packages to adopt packages which declare their platforms in the manifest. Doing so puts the brakes on adoption and ecosystem growth. For 'core' packages like this one availability constraints should be expressed on declarations rather than in the manifest.

This patch adds equivalent availability annotations to declarations across the package and removes platforms from the package manifest.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,15 +1,39 @@
 // swift-tools-version: 5.8
 
 import PackageDescription
+import CompilerPluginSupport
+
+// Availability Macros
+let availabilityTags = [_Availability("AsyncAlgorithms")]
+let versionNumbers = ["1.0"]
+
+// Availability Macro Utilities
+enum _OSAvailability: String {
+  // This should match the package's deployment target
+  case alwaysAvailable = "macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0"
+  // Use 10000 for future availability to avoid compiler magic around
+  // the 9999 version number but ensure it is greater than 9999
+  case future = "macOS 10000, iOS 10000, tvOS 10000, watchOS 10000"
+}
+
+struct _Availability {
+  let name: String
+  let osAvailability: _OSAvailability
+
+  init(_ name: String, availability: _OSAvailability = .alwaysAvailable) {
+    self.name = name
+    self.osAvailability = availability
+  }
+}
+
+let availabilityMacros: [SwiftSetting] = versionNumbers.flatMap { version in
+  availabilityTags.map {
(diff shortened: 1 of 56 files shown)


glbrntt:
@phausler note that Swift 5.6 and Swift 5.7 still use `platforms` per `Package@swift-5.7.swift` (`enableExperimentalFeature` and `CompilerPluginSupport` aren't available).

Package.swift
```diff
@@ -1,15 +1,39 @@
...
+import CompilerPluginSupport
+
+// Availability Macros
+let availabilityTags = [_Availability("AsyncAlgorithms")]
+let versionNumbers = ["1.0"]
+
+// Availability Macro Utilities
+enum _OSAvailability: String {
+  // This should match the package's deployment target
+  case alwaysAvailable = "macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0"
```
phausler, line 13: naming nit: this should probably be more so named `initialIntroduction` and not the "alwaysAvailable" since this is not something that has always been around.

Package.swift
```diff
@@ -1,15 +1,39 @@
...
+  // Use 10000 for future availability to avoid compiler magic around
+  // the 9999 version number but ensure it is greater than 9999
+  case future = "macOS 10000, iOS 10000, tvOS 10000, watchOS 10000"
```
phausler, line 16: it might be good to have a "pending" case for the 9999 versions

Package.swift
```diff
@@ -1,15 +1,39 @@
...
 
 import PackageDescription
+import CompilerPluginSupport
+
+// Availability Macros
+let availabilityTags = [_Availability("AsyncAlgorithms")]
+let versionNumbers = ["1.0"]
+
+// Availability Macro Utilities
+enum _OSAvailability: String {
```
phausler, line 11: naming nit: this doesn't need an underscore since this is not public API (it is restricted to the package)

Package.swift
```diff
@@ -1,15 +1,39 @@
...
+  // This should match the package's deployment target
+  case alwaysAvailable = "macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0"
+  // Use 10000 for future availability to avoid compiler magic around
+  // the 9999 version number but ensure it is greater than 9999
+  case future = "macOS 10000, iOS 10000, tvOS 10000, watchOS 10000"
+}
+
+struct _Availability {
```
phausler, line 19: ditto here on the underscore.

phausler (approved):

PR #357, opened 2025-07-28
[WIP] Draft pitch for share

phausler:


Evolution/NNNN-share.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Proposed solution
+
+AsyncAlgorithms will introduce a new extension function on AsyncSequence that will provide a shareable asynchronous sequence that will produce the same values upon iteration from multiple instances of it's AsyncIterator. Those iterations can take place in multiple isolations.
+
+When values from a differing isolation cannot be coalesced, the two options available are either awaiting (an exertion of back-pressure across the sequences) or buffering (an internal back-pressure to a buffer). Replaying the values from the beginning of the creation of the sequence is a distinctly different behavior that should be considered a different use case. This then leaves the behavioral characteristic of this particular operation of share as; sharing a buffer of values started from the initialization of a new iteration of the sequence. Control over that buffer should then have options to determine the behavior, similar to how AsyncStream allows that control. It should have options to be unbounded, buffering the oldest count of elements, or buffering the newest count of elements.
```
FranzBusch: I think we should expose both options as a configuration to share. So a user can choose to either exert back pressure or buffer depending on their use-case.

Evolution/NNNN-share.md
```diff
@@ -0,0 +1,178 @@
...
+
+It is critical to identify that this is one algorithm in the family of algorithms for sharing values. It should not attempt to solve all behavioral requirements but instead serve a common set of them that make cohesive sense together. This proposal is not mutually exclusive to the other algorithms in the sharing family.
+
+## Detailed design
+
+It is not just likely but perhaps a certainty that other algorithms will end up needing the same concept of a buffering policy beyond just AsyncStream and the new sharing mechanism. A new type in AsyncAlgorithms will be introduced to handle this. [^BufferingPolicy]
+
+```swift
+/// A strategy that handles exhaustion of a buffer’s capacity.
+public enum BufferingPolicy: Sendable {
```
FranzBusch: Extensible enums are not a feature for packages so this needs to become a struct with some `static`s instead similar to how we have done it in other algorithms such as `.buffer` in this repo. Additionally, there is little value here for a user to switch over this enum.

Evolution/NNNN-share.md
```diff
@@ -0,0 +1,178 @@
...
+  ///
+  /// This strategy enforces keeping at most the specified number of newest
+  /// values.
+  case bufferingNewest(Int)
+}
+```
+
+A new extension will be added to return a concrete type representing the share algorithm. This extension will take a buffering policy to identify how the buffer will be handled when iterations do not consume at the same rate.
+
+A new AsyncSequence type will be introduced that is explicitly marked as `Sendable`. This annotation identifies to the developer that this sequence can be shared and stored. Because the type is intended to be stored it cannot be returned by the extension as a `some AsyncSequence<Element, Failure> & Sendable` since that cannot be assigned to a stored property. Additionally the type of `AsyncShareSequence`, since indented to be stored, will act as a quasi erasing-barrier to the type information of previous sequences in the chain of algorithms in that it will only hold the generic information of the `Element` and `Failure` as part of it's public interface and not the "Base" asynchronous sequence it was created from.
```
FranzBusch: Why do we need to erase the base async sequence here? This is different than other algorithms and I'm sure this is going to have performance implications since we most likely need to box the base async sequence in an existential. 

I think using opaque return types is the _right_ solution here. Users can store that asynchronous sequence by just putting it behind an `any AsyncSequence<Element, Failure>` themselves. This leaves the choice to the user instead of forcing the existential on everyone.

Evolution/NNNN-share.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Runtime Behavior
+
+The construction of the `AsyncShareSequence` will initially construct a shared iteration reference. This means that all instances of the structure of the `AsyncShareSequence` will reference to the same iteration.
+
+Upon creation of the `Iterator` via `makeAsyncIterator` a new "side" will be constructed to identify the specific iterator interacting with the shared iteration. Then when next is invoked is where the first actual action takes place. 
+
+The next method will first checkout from a critical region the underlying AsyncIterator from the base. If that is successful (i.e. no other iteration sides have already checked it out) then it will invoke the next method of that iterator (forwarding in the actor isolation). If an element is produced then it enqueues the element to the shared buffer, checks in the iterator, adjusts the index in the buffer, and finds all pending continuations all in a shared critical region by a mutex. Then those continuations will be resumed with the given element.
```
FranzBusch: I don't think this is safe. In particular this section:

> invoke the next method of that iterator (forwarding in the actor isolation)

Async iterators are generally not-Sendable and you are proposing here to invoke the `next` method from potentially different isolation domains. While not concurrently you are proposing to do it sequentially. This might break some internal invariants of the base iterator. Furthermore, calling `next` directly from one of the iterators of `share` means that if that `next` call is canceled the base async sequence is cancelled. 

Instead I think we need to take a similar approach as we did in `merge` which is spawning an unstructured task on the first call to `next` and that task is going to be the only task to iterate the base asynchronous sequence. This upholds the Sendable constraints of the base and it correctly shields the base being cancelled by any of our share consumers.

Evolution/NNNN-share.md
```diff
@@ -0,0 +1,178 @@
...
+
+If no element is returned by the base iterator (signifying the terminal state);then the process is similar except it will instead mark the sequence as finished and resume with nil to any active continuations. Similarly with failures that will set the state as terminal but also store the error for further iteration points that need eventual termination.
+
+Then all sides are "drained" such that continuations are placed into the shared state and resumed when an element is available for that position.
+
+Practically this all means that a given iteration may be "behind" another and can eventually catch up (provided it is within the buffer limit).
```
FranzBusch: I think this is the reason why we need to provide the alternative variant as well that exerts back-pressure by the slowest consumer i.e. next on the base is only called once all current active iterators of share have consumed the last element. While that is most likely slower it guarantees the user that every iterator receives every element. On the other hand, with the currently proposed only buffer based solution there is no good way to set the size of the buffer that guarantees that no elements will ever be dropped.

Evolution/NNNN-share.md
```diff
@@ -0,0 +1,178 @@
...
+
+It is expected that this operator will be unlike other
+
+## Effect on API resilience
+
+This is an additive API and no existing systems are changed, however it will introduce a few new types that will need to be maintained as ABI interfaces. Since the intent of this is to provide a mechanism to store AsyncSequences to a shared context the type must be exposed as ABI (for type sizing).
+
+## Alternatives considered
+
+[^BufferingPolicy] It has been considered that this particular policy would be nested inside the `AsyncShareSequence` type. However since this seems to be something that will be useful for other types it makes sense to expose it as a top level type. However if it is determined that a general form of a  buffering policy would require additional behaviors this might be a debatable placement to move back to an interior type similar to AsyncStream.
```
FranzBusch: FWIW we already have a type that is pretty similar called `AsyncBufferSequencePolicy`. IMO we should stay somewhat consistent across the package.

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,455 @@
...
+  public func share(bufferingPolicy: AsyncBufferSequencePolicy = .unbounded) -> some AsyncSequence<Element, Failure> & Sendable {
+    return AsyncShareSequence(self, bufferingPolicy: bufferingPolicy)
+  }
+}
+
+@available(macOS 26.0, *)
+struct AsyncShareSequence<Base: AsyncSequence>: Sendable where Base.Element: Sendable {
+  final class Side {
+    struct State {
+      var continuaton: CheckedContinuation<Result<Element?, Failure>, Never>?
```
jamieQ: nit: 
```suggestion
      var continuation: CheckedContinuation<Result<Element?, Failure>, Never>?
```

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,455 @@
...
+      var cancelled = false
+      var limit: CheckedContinuation<Bool, Never>?
+      var demand: CheckedContinuation<Void, Never>?
+      
+      let storagePolicy: StoragePolicy
+      
+      init(_ base: Base, bufferingPolicy: AsyncBufferSequencePolicy) {
+        self.iteratingTask = .pending(base)
+        switch bufferingPolicy.policy {
+        case .bounded: self.storagePolicy = .unbounded
```
jamieQ, line 248: this mapping seems rather counter intuitive

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,455 @@
...
+        }
+        state.withLock { state in
+          precondition(state.iteratingTask.isStarting)
+          state.iteratingTask = .running(task)
+        }
+      }
+      let result: Result<Element?, Failure> = await withTaskCancellationHandler {
+        await withCheckedContinuation { continuation in
+          let (res, limitContinuation, demandContinuation, cancelled) = state.withLock { state -> (Result<Element?, Failure>?, CheckedContinuation<Bool, Never>?, CheckedContinuation<Void, Never>?, Bool) in
+            let side = state.sides[id]!
```
jamieQ: if the cancel handler races with this and wins will this crash?

phausler: hmm, I will have to dig into how to address that - good catch btw; I don't think it is impossible to address, just an oversight on my part

Sources/AsyncAlgorithms/AsyncShareSequence.swift
phausler: yea... when the buffering policy is set to bounded it means that the storage itself is unbounded but the limit is the bound; previously I split the external parameter of the policy out and this is a consequence of using the existing AsyncBufferSequencePolicy.

phausler: perhaps deserves a comment explaining why

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,455 @@
...
+          } catch {
+            emit(.failure(error as! Failure))
+          }
+        }
+        state.withLock { state in
+          precondition(state.iteratingTask.isStarting)
+          state.iteratingTask = .running(task)
+        }
+      }
+      let result: Result<Element?, Failure> = await withTaskCancellationHandler {
```
jamieQ: i haven't actually tested to verify, but i suspect the cancel handler body is currently subject to this [executor hopping issue](https://github.com/swiftlang/swift/pull/80753) if the isolated `actor` parameter isn't captured in the closure. i'm not sure it would be a correctness issue either way, but presumably the isolation should be forwarded so there's no need to risk hopping for no real reason – potentially just as a defensive measure to reduce the risk of surprising behavior.

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,455 @@
+import Synchronization
+
+@available(macOS 26.0, *)
+extension AsyncSequence where Element: Sendable {
```
FranzBusch: Doesn't the `Base` itself need to be `Sendable` here since we move it into an unstructured task?

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,455 @@
...
+  @available(macOS 26.0, *) // TODO: fix the availability for this to be defined as @available(AsyncAlgorithms 1.1, *)
+  public func share(bufferingPolicy: AsyncBufferSequencePolicy = .unbounded) -> some AsyncSequence<Element, Failure> & Sendable {
```
FranzBusch: I don't think this is a good nor safe default for all of Swift's use-cases. I would suggest we just always force users to explicitly pick a policy.

```suggestion
  public func share(bufferingPolicy: AsyncBufferSequencePolicy) -> some AsyncSequence<Element, Failure> & Sendable {
```

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,455 @@
...
+    deinit {
+      iteration.unregisterSide(id)
+    }
+    
+    func next(isolation actor: isolated (any Actor)?) async throws(Failure) -> Element? {
+      try await iteration.next(isolation: actor, id: id)
+    }
+  }
+  
+  final class Iteration: Sendable {
```
FranzBusch, line 183: Could we try to use the pattern established in `merge`, `zip` and `AsyncChannel` to write an explicit state machine with distinct methods and actions. This would make reviewing this code significantly easier.

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,455 @@
...
+    // this is the swapped state of transferring the base to the iterating task
+    // it does send the Base... but only one transfer
+    enum IteratingTask: @unchecked Sendable {
```
FranzBusch: This goes to my above comment. This `@unchecked Sendable` conformance is probably needed due to `Base` not being `Sendable`. In general, we should not need any unchecked conformances.

Sources/AsyncAlgorithms/AsyncShareSequence.swift
phausler: that kinda defeats the point.. I have a way around that but you aren't going to like it...

FranzBusch: I don't think it defeats the point. From what we discussed last year we essentially said every `AsyncSequence` should be `Sendable` but their iterators must not be `Sendable`. In fact, all `AsyncSequence`s in this package are `Sendable`. They just enforce runtime checks for a single iterator.

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,693 @@
...
+  ///
+  public func share(bufferingPolicy: AsyncBufferSequencePolicy = .bounded(1)) -> some AsyncSequence<Element, Failure> & Sendable {
+    // The iterator is transferred to the isolation of the iterating task
+    // this has to be done "unsafely" since we cannot annotate the transfer
+    // however since iterating an AsyncSequence types twice has been defined
+    // as invalid and one creation of the iterator is virtually a consuming
+    // operation so this is safe at runtime.
+    // The general principal of `.share()` is to provide a mecahnism for non-
+    // shared AsyncSequence types to be shared. The parlance for those is
+    // that the base AsyncSequence type is not Sendable. If the iterator
```
FranzBusch, line 78: I don't think this holds or is true really. Every async sequence that I know off right now is `Sendable`. This is inline with what we previously discussed where we said that `AsyncSequence`s are the blueprint and the iterators are the actual state. I really think we should constrain this extension to `Self: Sendable` here.

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,693 @@
...
+  /// ```
+  ///
+  /// - Parameter bufferingPolicy: The policy controlling how elements are enqueued to the shared buffer. Defaults to `.bounded(1)`.
+  ///   - `.bounded(n)`: Limits the buffer to `n` elements, applying backpressure to the source when that limit is reached
+  ///   - `.bufferingOldest(n)`: Keeps the oldest `n` elements, discarding newer ones when full
+  ///   - `.bufferingNewest(n)`: Keeps the newest `n` elements, discarding older ones when full
+  ///   - `.unbounded`: Allows unlimited buffering (use with caution)
+  ///
+  /// - Returns: A sendable async sequence that can be safely shared across multiple concurrent tasks.
+  ///
```
FranzBusch, line 67:
```suggestion
```

Sources/AsyncAlgorithms/AsyncShareSequence.swift
FranzBusch: I agree that this deserves a comment. I was very confused when I first read it

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,693 @@
...
+      let result: Result<Element?, Failure>
+      
+      func resume() {
+        continuation.resume(returning: result)
+      }
+    }
+    
+    func emit(_ result: Result<Element?, Failure>) {
+      let (resumptions, limitContinuation, demandContinuation, cancelled) = state.withLock { state -> ([Resumption], UnsafeContinuation<Bool, Never>?, UnsafeContinuation<Void, Never>?, Bool) in
+        var resumptions = [Resumption]()
```
FranzBusch, line 492: Instead of allocating a new array every time you might be better off re-using one

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -0,0 +1,693 @@
...
+        }
+      }
+      
+      // Internal state machine transitions for the emission of a given value.
+      //
+      // This method ensures the continuations are properly consumed when emitting values
+      // and returns those continuations for resumption.
+      //
+      // If no limit is specified it interprets that as an unbounded limit.
+      mutating func emit<T>(_ value: T, limit: Int?) -> (T, UnsafeContinuation<Bool, Never>?, UnsafeContinuation<Void, Never>?, Bool) {
```
FranzBusch: I find it really confusing that this method is called `emit` since it is called from the consumer from what I can tell. Can we call this `next` or something with `next` instead`?

Evolution/NNNN-share.md
```diff
@@ -0,0 +1,187 @@
+# Share
+
+## Introduction
+
+Many of the AsyncSequence adopting types only permit a one singular consumption. However there are many times that the same produced values are useful in more than one place. Out of that mechanism there are a few approaches to share, distribute, and broadcast those values. This proposal will focus on one concept; sharing. Sharing is where each consumption independently can make forward progress and  get the same values but do not replay from the beginning of time.
```
r3econ:
```suggestion
Many of the AsyncSequence adopting types only permit a one singular consumption. However there are many times that the same produced values are useful in more than one place. Out of that mechanism there are a few approaches to share, distribute, and broadcast those values. This proposal will focus on one concept; sharing. Sharing is where each consumption independently can make forward progress and get the same values but do not replay from the beginning of time.
```
Remove extra whitespace

Evolution/NNNN-share.md
```diff
@@ -0,0 +1,187 @@
...
+
+There are many potential usages for the sharing concept of AsyncSequences. 
+
+One such example is the case where a source of data as an asynchronous sequence needs to be consumed by updating UI, logging, and additionally  a network connection. This particular case does not matter on which uses but instead that those uses are independent of each other. It would not be expected for networking to block or delay the updates to UI, nor should logging. This example case also illustrates that the isolation of each side might be different and that some of the sides may not tolerate coalescing or dropping values.
+
+There are many other use cases that have been requested for this family of algorithms. Since the release of AsyncAlgorithms it has perhaps been the most popularly requested set of behaviors as additions to the package.
+
+## Proposed solution
+
+AsyncAlgorithms will introduce a new extension function on AsyncSequence that will provide a shareable asynchronous sequence that will produce the same values upon iteration from multiple instances of it's AsyncIterator. Those iterations can take place in multiple isolations.
```
r3econ:
```suggestion
AsyncAlgorithms will introduce a new extension function on AsyncSequence that will provide a shareable asynchronous sequence that will produce the same values upon iteration from multiple instances of its AsyncIterator. Those iterations can take place in multiple isolations.
```
Typo - possessive, not contraction

Evolution/NNNN-share.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Introduction
+
+Many of the AsyncSequence adopting types only permit a one singular consumption. However there are many times that the same produced values are useful in more than one place. Out of that mechanism there are a few approaches to share, distribute, and broadcast those values. This proposal will focus on one concept; sharing. Sharing is where each consumption independently can make forward progress and  get the same values but do not replay from the beginning of time.
+
+## Motivation
+
+There are many potential usages for the sharing concept of AsyncSequences. 
+
+One such example is the case where a source of data as an asynchronous sequence needs to be consumed by updating UI, logging, and additionally  a network connection. This particular case does not matter on which uses but instead that those uses are independent of each other. It would not be expected for networking to block or delay the updates to UI, nor should logging. This example case also illustrates that the isolation of each side might be different and that some of the sides may not tolerate coalescing or dropping values.
```
r3econ:
```suggestion
One such example is the case where a source of data as an asynchronous sequence needs to be consumed by updating UI, logging, and additionally a network connection. This particular case does not matter on which uses but instead that those uses are independent of each other. It would not be expected for networking to block or delay the updates to UI, nor should logging. This example case also illustrates that the isolation of each side might be different and that some of the sides may not tolerate coalescing or dropping values.
```
Remove extra whitespace

Evolution/NNNN-share.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Runtime Behavior
+
+The runtime behaviors fall into a few categories; ordering, iteration isolation, cancellation, and lifetimes. To understand the beahviors there are a terms useful to define. Each creation of the AsyncIterator of the sequence and invocation of next will be referred to a side of the share iteration. The back pressure to the system to fetch a new element or termination is refered to as demand. The limit which is the pending gate for awaiting until the buffer has been serviced used for the `AsyncBufferSequencePolicy.bounded(_ : Int)` policy. The last special definition is that of the extent which is specifically in this case the lifetime of the asynchronous sequence itself.
```
r3econ:
```suggestion
The runtime behaviors fall into a few categories; ordering, iteration isolation, cancellation, and lifetimes. To understand the behaviors there are terms useful to define. Each creation of the AsyncIterator of the sequence and invocation of next will be referred to a side of the share iteration. The back pressure to the system to fetch a new element or termination is referred to as demand. The limit which is the pending gate for awaiting until the buffer has been serviced used for the `AsyncBufferSequencePolicy.bounded(_ : Int)` policy. The last special definition is that of the extent which is specifically in this case the lifetime of the asynchronous sequence itself.
```
Fix typos

r3econ (approved):

phausler:
@r3econ swift-async-algorithms is a separate package from Swift itself so this must means that folks can use the main branch for this new feature (provided they are using swift 6.0 or greater)

I am however considering tagging a minor release update to take some of the other fixes here for the primary releases before tagging a release with share.

It would generally be considered that the main branch is suitable for testing out and gathering feedback; e.g. if you have an issue with it please file a bug and tag me so it can be fixed asap.

PR #362, opened 2025-08-20
Update the project to address concurrency failures for 6.2 build modes.

phausler:
When updating to the latest language mode there are a ton of warnings that became errors around Swift Concurrency. This addresses all of the errors so far (and caught a subtile compiler bug for sendable types within the buffer algorithm).

This still has a few warnings around Sendable meta types.

e.g.

```
AsyncChunksOfCountOrSignalSequence.swift:169:33: warning: capture of non-Sendable type 'Collected.Type' in an isolated closure [#SendableMetatypes]
167 |     return Iterator(
168 |       iterator: merge(
169 |         chain(base.map { Either.element($0) }, [.terminal].async),
    |                                 `- warning: capture of non-Sendable type 'Collected.Type' in an isolated closure [#SendableMetatypes]
170 |         signal.map { _ in Either.signal }
171 |       ).makeAsyncIterator(),
```

These are relatively prolific but likely need to be addressed by restricting around a where clause requiring `SendableMetatype` https://docs.swift.org/compiler/documentation/diagnostics/sendable-metatypes/

### DIFF
--- a/Sources/AsyncSequenceValidation/TaskDriver.swift
+++ b/Sources/AsyncSequenceValidation/TaskDriver.swift
@@ -53,7 +53,7 @@ func start_thread(_ raw: UnsafeMutableRawPointer) -> UnsafeMutableRawPointer {
 final class TaskDriver: Sendable {
   let work: @Sendable (TaskDriver) -> Void
   let queue: WorkQueue
-  #if canImport(Darwin)
+  #if canImport(Darwin) || canImport(wasi_pthread)
   nonisolated(unsafe) var thread: pthread_t?
   #elseif canImport(Glibc) || canImport(Musl) || canImport(Bionic)
   nonisolated(unsafe) var thread = pthread_t()


Sources/AsyncAlgorithms/Buffer/AsyncBufferSequence.swift
```diff
@@ -121,9 +121,9 @@ public struct AsyncBufferSequence<Base: AsyncSequence & Sendable>: AsyncSequence
         self.storageType = .transparent(iterator)
         return element
       case .bounded(let storage):
-        return try await storage.next()?._rethrowGet()
+        return try await storage.next().wrapped?._rethrowGet()
       case .unbounded(let storage):
-        return try await storage.next()?._rethrowGet()
+        return try await storage.next().wrapped?._rethrowGet()
```
FranzBusch, line 134: Can we leave a comment here why the unsafe transfer is needed and why it is sound?

PR #364, opened 2025-09-05
Retry & Backoff

ph1ps:
No docs nor a pitch nor unit tests, yet. But I am looking for early feedback on the interface of the proposed algorithm.
Usage would look something like this:
``` swift
// Example of when backoff is calculated by a client
let rng = SystemRandomNumberGenerator() // or a seeded rng for unit tests
var backoff = Backoff
  .exponential(factor: 2, initial: .seconds(5))
  .maximum(.seconds(100))
  .fullJitter(using: rng)
  .minimum(.seconds(2))

let response = try await retry(maxAttempts: 5) {
  try await URLSession.shared.data(from: url)
} strategy: { error in
  return .backoff(backoff.nextDuration())
}
```

```swift
// Example of when backoff is returned by a server
let response = try await retry(maxAttempts: 5) {
  let (data, response) = try await URLSession.shared.data(from: url)
  if
    let response = response as? HTTPURLResponse,
    response.statusCode == 429,
    let retryAfter = response.value(forHTTPHeaderField: "Retry-After")
  {
   throw TooManyRequestsError(retryAfter: Double(retryAfter)!)
  }
  return (data, response)
} strategy: { error in
  if let error = error as? TooManyRequestsError {
    return .backoff(.seconds(error.retryAfter))
  } else {
    return .stop
  }
}
```
(Of course nothing in this algorithm is limited to networking, in fact I have used this multiple times with Bluetooth implementations, but network code is more widely used, so I used this in my examples...)

Retry on an `AsyncSequence` could be a future direction, although I have not yet found a use case for this.

Thank you.

### DIFF
--- a/Sources/AsyncAlgorithms/Retry/Backoff.swift
+++ b/Sources/AsyncAlgorithms/Retry/Backoff.swift
@@ -1,84 +1,144 @@
 #if compiler(>=6.2)
 /// A protocol for defining backoff strategies that generate delays between retry attempts.
 ///
-/// Each call to `nextDuration()` returns the delay for the next retry attempt. Strategies are
-/// naturally stateful. For instance, they may track the number of invocations or the previously
-/// returned duration to calculate the next delay.
+/// A `BackoffStrategy` represents an immutable configuration for generating delay durations.
+/// To produce actual delay values, call `makeIterator()` to create a `BackoffIterator`.
+/// This separation allows strategies to be `Sendable` and reusable, while iterators manage
+/// the mutable state for generating successive delays.
 ///
 /// ## Example
 ///
 /// ```swift
-/// var strategy = Backoff.exponential(factor: 2, initial: .milliseconds(100))
-/// strategy.nextDuration() // 100ms
-/// strategy.nextDuration() // 200ms
-/// strategy.nextDuration() // 400ms
+/// let strategy = Backoff.exponential(factor: 2, initial: .milliseconds(100))
+/// var iterator = strategy.makeIterator()
+/// iterator.nextDuration() // 100ms
+/// iterator.nextDuration() // 200ms
+/// iterator.nextDuration() // 400ms
 /// ```
 @available(AsyncAlgorithms 1.1, *)
 public protocol BackoffStrategy<Duration> {
+  associatedtype Iterator: BackoffIterator
+  associatedtype Duration: DurationProtocol where Duration == Iterator.Duration
+  func makeIterator() -> Iterator
+}
+
+/// A protocol for stateful iteration over backoff delay durations.
+///
+/// A `BackoffIterator` is created from a `BackoffStrategy` via `makeIterator()`.
+/// Each call to `nextDuration()` returns the delay for the next retry attempt.
+/// Iterators are stateful; they may track the number of invocations or the
+/// previously returned duration to calculate the next delay.
+@available(AsyncAlgorithms 1.1, *)
+public protocol BackoffIterator {
   associatedtype Duration: DurationProtocol
   mutating func nextDuration() -> Duration
+  mutating func nextDuration(using generator: inout some RandomNumberGenerator) -> Duration
 }
 
 @available(AsyncAlgorithms 1.1, *)
-@usableFromInline struct ConstantBackoffStrategy<Duration: DurationProtocol>: BackoffStrategy {
+extension BackoffIterator {
+  /// Default implementation that ignores the random number generator and
+  /// calls ``nextDuration()``.
+  ///
+  /// Override this method in iterators that use randomization (such as jitter)
+  /// to use the provided generator instead of the system default.
+  public mutating func nextDuration(using generator: inout some RandomNumberGenerator) -> Duration {
+    nextDuration()
+  }
+}
+
+@available(AsyncAlgorithms 1.1, *)
+@usableFromInline struct ConstantBackoffStrategy<Duration: DurationProtocol>: BackoffStrategy, Sendable {
   @usableFromInline let constant: Duration
   @usableFromInline init(constant: Duration) {
     precondition(constant >= .zero, "Constant must be greater than or equal to 0")
     self.constant = constant
   }
-  @inlinable func nextDuration() -> Duration {
-    return constant
+  @inlinable func makeIterator() -> Iterator {
+    return Iterator(constant: constant)
+  }
+  @usableFromInline struct Iterator: BackoffIterator {
+    @usableFromInline let constant: Duration
+    @usableFromInline init(constant: Duration) {
+      self.constant = constant
+    }
+    @inlinable @inline(__always) func nextDuration() -> Duration {
+      return constant
+    }
   }
 }
 
 @available(AsyncAlgorithms 1.1, *)
-@usableFromInline struct LinearBackoffStrategy: BackoffStrategy {
-  @usableFromInline var current: Duration
+@usableFromInline struct LinearBackoffStrategy: BackoffStrategy, Sendable {
+  @usableFromInline let initial: Duration
   @usableFromInline let increment: Duration
-  @usableFromInline var hasOverflown = false
   @usableFromInline init(increment: Duration, initial: Duration) {
     precondition(initial >= .zero, "Initial must be greater than or equal to 0")
     precondition(increment >= .zero, "Increment must be greater than or equal to 0")
-    self.current = initial
+    self.initial = initial
     self.increment = increment
   }
-  @inlinable mutating func nextDuration() -> Duration {
-    if hasOverflown {
-      return Duration(attoseconds: .max)
-    } else {
-      let (next, hasOverflown) = current.attoseconds.addingReportingOverflow(increment.attoseconds)
+  @inlinable func makeIterator() -> Iterator {
+    return Iterator(current: initial, increment: increment)
+  }
+  @usableFromInline struct Iterator: BackoffIterator {
+    @usableFromInline var current: Duration
+    @usableFromInline let increment: Duration
+    @usableFromInline var hasOverflown = false
+    @usableFromInline init(current: Duration, increment: Duration) {
+      self.current = current
+      self.increment = increment
+    }
+    @inlinable @inline(__always) mutating func nextDuration() -> Duration {
       if hasOverflown {
-        self.hasOverflown = true
-        return nextDuration()
+        return Duration(attoseconds: .max)
       } else {
-        defer { current = Duration(attoseconds: next) }
-        return current
+        let (next, hasOverflown) = current.attoseconds.addingReportingOverflow(increment.attoseconds)
+        if hasOverflown {
+          self.hasOverflown = true
+          return Duration(attoseconds: .max)
+        } else {
+          defer { current = Duration(attoseconds: next) }
+          return current
+        }
       }
     }
   }
 }
 
 @available(AsyncAlgorithms 1.1, *)
-@usableFromInline struct ExponentialBackoffStrategy: BackoffStrategy {
-  @usableFromInline var current: Duration
+@usableFromInline struct ExponentialBackoffStrategy: BackoffStrategy, Sendable {
+  @usableFromInline let initial: Duration
   @usableFromInline let factor: Int128
-  @usableFromInline var hasOverflown = false
   @usableFromInline init(factor: Int128, initial: Duration) {
     precondition(initial >= .zero, "Initial must be greater than or equal to 0")
-    self.current = initial
+    precondition(factor >= 1, "Factor must be greater than or equal to 1")
+    self.initial = initial
     self.factor = factor
   }
-  @inlinable mutating func nextDuration() -> Duration {
-    if hasOverflown {
-      return Duration(attoseconds: .max)
-    } else {
-      let (next, hasOverflown) = current.attoseconds.multipliedReportingOverflow(by: factor)
+  @inlinable func makeIterator() -> Iterator {
+    return Iterator(current: initial, factor: factor)
+  }
+  @usableFromInline struct Iterator: BackoffIterator {
+    @usableFromInline var current: Duration
+    @usableFromInline let factor: Int128
+    @usableFromInline var hasOverflown = false
+    @usableFromInline init(current: Duration, factor: Int128) {
+      self.current = current
+      self.factor = factor
+    }
+    @inlinable @inline(__always) mutating func nextDuration() -> Duration {
       if hasOverflown {
-        self.hasOverflown = true
-        return nextDuration()
+        return Duration(attoseconds: .max)
       } else {
-        defer { current = Duration(attoseconds: next) }
-        return current
+        let (next, hasOverflown) = current.attoseconds.multipliedReportingOverflow(by: factor)
+        if hasOverflown {
+          self.hasOverflown = true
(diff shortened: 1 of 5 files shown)


Sources/_CPowSupport/include/_CPowSupport.h
```diff
@@ -0,0 +1,3 @@
+static inline __attribute__((__always_inline__)) double pow(double x, double y) {
```
phausler: you could avoid the need for an extra module by enabling the Extern experimental feature and using this from swift:

```
@_extern(c, "__builtin_pow")
func pow(_ x: Double, _ y: Double) -> Double
```

Sources/AsyncAlgorithms/Retry/Backoff.swift
```diff
@@ -0,0 +1,198 @@
...
+@available(iOS 16.0, macCatalyst 16.0, macOS 13.0, tvOS 16.0, visionOS 1.0, watchOS 9.0, *)
+public protocol BackoffStrategy<Duration> {
+  associatedtype Duration: DurationProtocol
+  mutating func duration(_ attempt: Int) -> Duration
+  mutating func duration(_ attempt: Int, using generator: inout some RandomNumberGenerator) -> Duration
+}
+
+@available(iOS 16.0, macCatalyst 16.0, macOS 13.0, tvOS 16.0, visionOS 1.0, watchOS 9.0, *)
+extension BackoffStrategy {
+  @inlinable public mutating func duration(_ attempt: Int, using generator: inout some RandomNumberGenerator) -> Duration {
```
phausler: shouldn't this be the other way around? e.g. the attempt only version has a default implementation of passing a default RNG? because it isn't required that an implementation actually uses that RNG...

Sources/AsyncAlgorithms/Retry/Backoff.swift
```diff
@@ -0,0 +1,198 @@
...
+    self.b = b
+  }
+  @inlinable func duration(_ attempt: Int) -> Duration {
+    return a * attempt + b
+  }
+}
+
+@available(iOS 16.0, macCatalyst 16.0, macOS 13.0, tvOS 16.0, visionOS 1.0, watchOS 9.0, *)
+@usableFromInline struct ExponentialBackoffStrategy: BackoffStrategy {
+  @usableFromInline let a: Duration
```
phausler: amount and factor might be better names than a and b imho. Naming is good even for internal implementations ;)

Sources/AsyncAlgorithms/Retry/Backoff.swift
```diff
@@ -0,0 +1,198 @@
...
+  @usableFromInline let b: Double
+  @usableFromInline init(a: Duration, b: Double) {
+    self.a = a
+    self.b = b
+  }
+  @inlinable func duration(_ attempt: Int) -> Duration {
+    return a * pow(b, Double(attempt))
```
phausler: since this is mutating this could be done w/o pow iiuc.

Sources/AsyncAlgorithms/Retry/Backoff.swift
```diff
@@ -0,0 +1,198 @@
...
+    self.init(_high: attoseconds._high, low: attoseconds._low)
+  }
+}
+#endif
+
+@available(iOS 16.0, macCatalyst 16.0, macOS 13.0, tvOS 16.0, visionOS 1.0, watchOS 9.0, *)
+public protocol BackoffStrategy<Duration> {
+  associatedtype Duration: DurationProtocol
+  mutating func duration(_ attempt: Int) -> Duration
+  mutating func duration(_ attempt: Int, using generator: inout some RandomNumberGenerator) -> Duration
```
phausler: it seems the RNG is only used for the jitter variant; im not sure it is really a protocol requirement type thing and more-so a specific instance construction thing.

Could the attempt index also be entrusted to instead be tracked by the backoff strategy itself? That way the indexing of that doesn't need to be externally stored and the duration is just the responsibility of the strategy to figure out how to determine what the next value is?

Sources/AsyncAlgorithms/Retry/Retry.swift
```diff
@@ -0,0 +1,51 @@
+@available(iOS 16.0, macCatalyst 16.0, macOS 13.0, tvOS 16.0, visionOS 1.0, watchOS 9.0, *)
+public enum RetryStrategy<Duration: DurationProtocol> {
```
phausler: this would likely need to be a structure backed by an enum due to the package layout we have

Sources/AsyncAlgorithms/Retry/Backoff.swift
ph1ps: IIUC that would not work with exotic cases like these:

```swift
var constantBackoff = Backoff.constant(.seconds(1))
var expontentialBackoff = Backoff.exponential(initial: .seconds(5))

let response = try await retry {
  try await URLSession.shared.data(from: url)
} strategy: { attempt, error in
  if Bool.random() {
    return .backoff(exponentialBackoff.duration(attempt))
  } else {
    return .backoff(constantBackoff.duration(attempt))
  }
}
```

Sources/AsyncAlgorithms/Retry/Backoff.swift
ph1ps: I experimented with this a little, that was my intent behind this question: https://forums.swift.org/t/lifetime-dependencies-and-pointers/81943.

IMO it would've been really elegant to have the rng be part of the specific jitter strategies, but I couldn't find a way to not make a copy and keep the inout-ness of rng.

Sources/AsyncAlgorithms/Retry/Backoff.swift
phausler: we don't need to support all uses cases especially if it means that the more common cases suffer from extra complexity.

Sources/AsyncAlgorithms/Retry/Backoff.swift
```diff
@@ -0,0 +1,184 @@
...
+    return next
+  }
+}
+
+@available(iOS 16.0, macCatalyst 16.0, macOS 13.0, tvOS 16.0, visionOS 1.0, watchOS 9.0, *)
+public enum Backoff {
+  @inlinable public static func constant<Duration: DurationProtocol>(_ constant: Duration) -> some BackoffStrategy<Duration> {
+    return ConstantBackoffStrategy(constant: constant)
+  }
+  @inlinable public static func constant(_ constant: Duration) -> some BackoffStrategy<Duration> {
```
FranzBusch: Duration is conforming to `Duration` protocol so do we need these concrete overloads?

Sources/AsyncAlgorithms/Retry/Retry.swift
```diff
@@ -0,0 +1,61 @@
...
+    return .init(strategy: .stop)
+  }
+  @inlinable public static func backoff(_ duration: Duration) -> Self {
+    return .init(strategy: .backoff(duration))
+  }
+}
+
+@available(iOS 16.0, macCatalyst 16.0, macOS 13.0, tvOS 16.0, visionOS 1.0, watchOS 9.0, *)
+@inlinable public func retry<Result, ErrorType, ClockType>(
+  maxAttempts: Int = 3,
```
FranzBusch: This seems like an arbitrary default to me. We are probably better off without a default.

Sources/AsyncAlgorithms/Retry/Retry.swift
```diff
@@ -0,0 +1,61 @@
...
+  tolerance: ClockType.Instant.Duration? = nil,
+  clock: ClockType,
+  isolation: isolated (any Actor)? = #isolation,
+  operation: () async throws(ErrorType) -> sending Result,
```
FranzBusch: Why is the result `sending` here?

Sources/AsyncAlgorithms/Retry/Retry.swift
```diff
@@ -0,0 +1,61 @@
...
+  strategy: (ErrorType) -> RetryStrategy<ClockType.Instant.Duration> = { _ in .backoff(.zero) }
```
FranzBusch: What if we make `RetryStrategy` hold this closure directly and become generic itself? That allows for easier sharing of strategies.

Sources/AsyncAlgorithms/Retry/Retry.swift
```diff
@@ -0,0 +1,61 @@
...
+) async throws -> Result where ClockType: Clock, ErrorType: Error {
+  precondition(maxAttempts > 0, "Must have at least one attempt")
+  for _ in 0..<maxAttempts - 1 {
+    do {
+      return try await operation()
+    } catch where Task.isCancelled {
```
FranzBusch: I don't think we should be checking for task cancellation here. This should be up to the strategy.

Sources/AsyncAlgorithms/Retry/Retry.swift
```diff
@@ -0,0 +1,61 @@
...
+public struct RetryStrategy<Duration: DurationProtocol> {
+  @usableFromInline enum Strategy {
+    case backoff(Duration)
+    case stop
+  }
+  @usableFromInline let strategy: Strategy
+  @usableFromInline init(strategy: Strategy) {
+    self.strategy = strategy
+  }
+  @inlinable public static var stop: Self {
```
FranzBusch, line 14: Is this really a retry strategy? It has the same behaviour as max attempts = 1 right?

Sources/AsyncAlgorithms/Retry/Backoff.swift
ph1ps: You are right, however in practice I found it useful to have these overloads to make it a bit easier to write code like this:
```
var strategy = Backoff.constant(.milliseconds(5))
```
... whereas without the overloads you have to be more specific:
```
var strategy = Backoff.constant(Duration.milliseconds(5))
```
This matches also the overload with the default clock of `retry` which preselects `ContinuousClock` which duration is `Duration`.

Felt like a QoL to me, but I think I'd be open to get rid of them, if we say this is confusing.

FranzBusch: Ah yeah that's a good point.

Sources/AsyncAlgorithms/Retry/Retry.swift
ph1ps: I don't know if this was a bug in Swift 6 & 6.1, but it was required to have pre 6.2: https://godbolt.org/z/c9K7jEjcq

Sources/AsyncAlgorithms/Retry/Retry.swift
ph1ps: It has, but if you consider this example:
```swift
let response = try await retry {
  let (data, response) = try await URLSession.shared.data(from: url)
  if
    let response = response as? HTTPURLResponse,
    response.statusCode == 429,
    let retryAfter = response.value(forHTTPHeaderField: "Retry-After")
  {
   throw TooManyRequestsError(retryAfter: Double(retryAfter)!)
  }
  return (data, response)
} strategy: { error in
  if let error = error as? TooManyRequestsError {
    return .backoff(.seconds(error.retryAfter))
  } else {
    return .stop
  }
}
```
... you need some dynamic way of saying, enough, stop retrying. You'd use this in retry/backoff unrelated errors, otherwise it'd keep retrying even though you get a 404 or something, which will probably not change if you try again a couple of times. You'll end up just wasting resources.

Sources/AsyncAlgorithms/Retry/Retry.swift
FranzBusch: Yeah this was a bug before which means we might only be able to write this algo in 6.2

Sources/AsyncAlgorithms/Retry/Retry.swift
ph1ps: That is an interesting use case. I will investigate.

Sources/AsyncAlgorithms/Retry/Retry.swift
czechboy0: Maybe call it something like `Action`, `NextAction`, `RetryAction`? As it's really an instruction of what to do next, rather than a strategy that you'd know in advance.

I agree that this needs to be controlled dynamically, as HTTP/gRPC both have a clear distinction between "retryable" and "non-retryable" errors, and it's a very bad practice to retry non-retryable errors (especially the 429 above).

Sources/AsyncAlgorithms/Retry/Retry.swift
ph1ps: You are right. Maybe it'd be worth  backporting with `sending` with the downside of it being slightly more restrictive, but I will guard >= 6.2, for now.

ph1ps:
@phausler 
I think this is ready for pitching. What's the process of evolution with this project? Do I post in `Related Projects -> Async Algorithms` or do I post in `Evolution -> Pitches`? Is there a review period, or is the pitch thread the review thread?

Sources/AsyncAlgorithms/Retry/Backoff.swift
```diff
@@ -0,0 +1,347 @@
...
+@usableFromInline struct ExponentialBackoffStrategy<Duration: DurationProtocol>: BackoffStrategy {
+  @usableFromInline var current: Duration
+  @usableFromInline let factor: Int
+  @usableFromInline init(factor: Int, initial: Duration) {
+    precondition(initial >= .zero, "Initial must be greater than or equal to 0")
+    self.current = initial
+    self.factor = factor
+  }
+  @inlinable mutating func nextDuration() -> Duration {
+    defer { current *= factor }
```
bobergj: Looks like this multiplication can overflow when retrying many times. I just saw it happen with `.exponential(factor: 2, initial: .seconds(5))`. This should check for overflows and cap the value to the max of the duration type.

ph1ps: @bobergj Although this can overflow at some point it is not realistic, in my opinion.

The last duration before the multiplication (with factor 2 and initial 5 seconds) overflows is 1.46 trillion years, if I did the math correct.

It is generally a good idea to cap exponential backoff at a maximum which makes sense for your use case. 

Would you mind sharing what you were trying to do? Maybe I should add a recommendation to the docs of exponential backoff that it should be combined with a .max modifier...

bobergj: Sorry for the incomplete report. On iOS < 18 I am doing this:
```
     Backoff
         .exponential(factor: 2, initial: .seconds(5))
         .maximum(.seconds(120))
  ```
It turns out the `.maximum` is part of the issue. Yes, with a pure exponential backoff, you may have to wait a trillion years. But with this setup the exponential keeps growing, while the waiting time is much shorter due to the .maximum. This overflows after about 65 calls to `nextDuration`, which is after a total retry time of just over an hour.

ph1ps: @bobergj Yikes, now I get it. Thanks for the report.

I'll either implement what you suggested or the max-modifier should  stop calling its base backoff strategy once maximum was reached.

ph1ps: @bobergj I've updated the implementation and proposal: https://forums.swift.org/t/pitch-retry-backoff/82483/2. Thanks again.

Issue #367, opened 2025-09-12
New 1.1 operators like .share() won't be available on older Apple platforms (eg: iOS <18)


stefanomondino:
I was watching closely for months the development of the new `.share()` operator that [just got merged](https://github.com/apple/swift-async-algorithms/pull/357) because our apps are lacking an effective way to "mimic" Combine's Publishers (like exposing a "stream" variable that can be consumed in multiple places of the app).

However, I didn't fully realized at first that this operator (and more generally speaking, every new one that will be released as 1.1) will have a **very strict requirement of iOS 18.0+** (and same for tvOS, macOS, etc - OS versions from 2024).

This is blocking us from adopting it in all of our apps - it's common to support current OS minus 2 as business requirement for new apps and - as I'm writing - this means a min version of iOS 16.0 (iOS 26 will be officially released next week). And this applies only to new apps, because for already existing one it's really unlikely to drop support to older devices unless really needed (and proven).

The adoption of `.share()` is important, because it's preventing a proper use of custom sequences in large scale architectures (if I'm exposing a sequence with a variable I get no formal guarantee that some other Task is consuming it). And @phausler already stated the same in the share pitch + the massive work on the implementation - that's actually where I've learned a lot of things on this subject.

I'd like to better understand how we developers are supposed to proceed in these cases, because for at least a couple of years we won't be able to adopt these new features. The alternatives I see:
1) we stick to Combine -> Not 100% compatible with strict concurrency, and the point of it all is to get rid of it.
2) the implementations get backported to older versions as well, by creating custom alternatives to those methods (example: share operatore uses heavily .withLock from Mutex, which is probably the reason why it's getting constrained to iOS 18. One could add a conditional compilation that uses... something else on older Apple platforms instead).  
3) we implement custom solutions in our projects to allow share/broadcast/whatever we need -> risky, but seems the only viable solution without interventions from the official repo maintainers.

[here](https://github.com/apple/swift-async-algorithms/pull/359#issuecomment-3285847360) we had a brief discussion with @phausler about it but I don't understand how Swift 6.2 is involved compared to os version.

Also it's not clear to me how new features will be added to this repository - is it 1.1 entirely going to be iOS 18+ compatible? With the new availability macros I'm a little bit lost. Will it be possible to mark a new feature as 1.0 available even after 1.1 is out?

[...]

PR #369, opened 2025-09-22
[WIP] 1.0 compatibility for share operator

stefanomondino:
fixes: #367 

As discussed in the issue with @FranzBusch, this is an attempt to allow the new `share` operator to be available on older Apple platforms (before iOS 18.0 and similar), also known as 1.0 compatibility.

> Note: when referring to iOS *** I mean of course all the Apple platforms released in the same year

This comes with a couple of tradeoffs

1. `Mutex` had to be replaced by `ManagedCriticalState`, as it's not available on 1.0
2. `Failure` generic type for errors had to be replaced with a non-generic `Swift.Error` type.
3. (really minor in my opinion): tests were not compiling on iOS simulators via Xcode with the `.milliseconds(10)` definitions (iOS 15.0 min compatibility),  and I had to switch to nanoseconds. Using older iOS versions was the quickest thing to do on my end to test this scenario.

When ran on iOS 17.5 unfortunately `ManagedCriticalState.withLock` is crashing for some unknown-to-me reason - I know @phausler already encountered this and I didn't fully understand if there's a possible alternative solution / workaround or not.

I'm opening this here as a starting point.

Also, I was thinking if something [this backport of Mutex](https://gist.github.com/swhitty/571deb25d84c1954a7a01aafa661496e) could help, in terms of approach:
This is only iOS 16.0+ compatible (not 13.0) but this would be a huge step forward for us developers, since it's way easier to have business direction to accept 16.0 as minimum support, compared to 18.0.

This would probably require the entire library to rename 1.1 into 1.2, having 1.1 to be "intermediately" compatible with iOS 16.0 and 1.2 starting from 18.0 - I perfectly understand this is a huge change but I really think that including share (at least this single one!) it's critical for this package adoption.

### DIFF
--- a/Sources/AsyncAlgorithms/AsyncShareSequence.swift
+++ b/Sources/AsyncAlgorithms/AsyncShareSequence.swift
@@ -9,14 +9,12 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if compiler(>=6.2)
-
 import Synchronization
 import DequeModule
 
-@available(AsyncAlgorithms 1.1, *)
+@available(AsyncAlgorithms 1.0, *)
 extension AsyncSequence
-where Element: Sendable, Self: SendableMetatype, AsyncIterator: SendableMetatype {
+where Element: Sendable, Self: _SendableMetatype, AsyncIterator: _SendableMetatype {
   /// Creates a shared async sequence that allows multiple concurrent iterations over a single source.
   ///
   /// The `share` method transforms an async sequence into a shareable sequence that can be safely
@@ -67,7 +65,7 @@ where Element: Sendable, Self: SendableMetatype, AsyncIterator: SendableMetatype
   ///
   public func share(
     bufferingPolicy: AsyncBufferSequencePolicy = .bounded(1)
-  ) -> some AsyncSequence<Element, Failure> & Sendable {
+  ) -> AsyncShareSequence<Self> {
     // The iterator is transferred to the isolation of the iterating task
     // this has to be done "unsafely" since we cannot annotate the transfer
     // however since iterating an AsyncSequence types twice has been defined
@@ -114,9 +112,9 @@ where Element: Sendable, Self: SendableMetatype, AsyncIterator: SendableMetatype
 //
 // This type is typically not used directly; instead, use the `share()` method on any
 // async sequence that meets the sendability requirements.
-@available(AsyncAlgorithms 1.1, *)
-struct AsyncShareSequence<Base: AsyncSequence>: Sendable
-where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: SendableMetatype {
+@available(AsyncAlgorithms 1.0, *)
+public struct AsyncShareSequence<Base: AsyncSequence>: Sendable
+where Base.Element: Sendable, Base: _SendableMetatype, Base.AsyncIterator: _SendableMetatype {
   // Represents a single consumer's connection to the shared sequence.
   //
   // Each iterator of the shared sequence creates its own `Side` instance, which tracks
@@ -135,7 +133,7 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
     // - `continuation`: The continuation waiting for the next element (nil if not waiting)
     // - `position`: The consumer's current position in the shared buffer
     struct State {
-      var continuation: UnsafeContinuation<Result<Element?, Failure>, Never>?
+        var continuation: UnsafeContinuation<Result<Base.Element?, Error>, Never>?
       var position = 0
 
       // Creates a new state with the position adjusted by the given offset.
@@ -162,7 +160,7 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
       iteration.unregisterSide(id)
     }
 
-    func next(isolation actor: isolated (any Actor)?) async throws(Failure) -> Element? {
+    func next(isolation actor: isolated (any Actor)?) async throws -> Base.Element? {
       try await iteration.next(isolation: actor, id: id)
     }
   }
@@ -230,9 +228,9 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
       var generation = 0
       var sides = [Int: Side.State]()
       var iteratingTask: IteratingTask
-      private(set) var buffer = Deque<Element>()
+        private(set) var buffer = Deque<Base.Element>()
       private(set) var finished = false
-      private(set) var failure: Failure?
+      private(set) var failure: Error?
       var cancelled = false
       var limit: UnsafeContinuation<Bool, Never>?
       var demand: UnsafeContinuation<Void, Never>?
@@ -311,7 +309,7 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
       //   **Buffering Newest**: Appends if under the limit, otherwise removes the oldest and appends
       //
       // - Parameter element: The element to add to the buffer
-      mutating func enqueue(_ element: Element) {
+        mutating func enqueue(_ element: Base.Element) {
         let count = buffer.count
 
         switch storagePolicy {
@@ -335,20 +333,20 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
         finished = true
       }
 
-      mutating func fail(_ error: Failure) {
+      mutating func fail(_ error: Error) {
         finished = true
         failure = error
       }
     }
 
-    let state: Mutex<State>
+    let state: ManagedCriticalState<State>
     let limit: Int?
 
     init(
       _ iteratorFactory: @escaping @Sendable () -> sending Base.AsyncIterator,
       bufferingPolicy: AsyncBufferSequencePolicy
     ) {
-      state = Mutex(State(iteratorFactory, bufferingPolicy: bufferingPolicy))
+      state = ManagedCriticalState(State(iteratorFactory, bufferingPolicy: bufferingPolicy))
       switch bufferingPolicy.policy {
       case .bounded(let limit):
         self.limit = limit
@@ -478,15 +476,15 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
     }
 
     struct Resumption {
-      let continuation: UnsafeContinuation<Result<Element?, Failure>, Never>
-      let result: Result<Element?, Failure>
+        let continuation: UnsafeContinuation<Result<Base.Element?, Error>, Never>
+        let result: Result<Base.Element?, Error>
 
       func resume() {
         continuation.resume(returning: result)
       }
     }
 
-    func emit(_ result: Result<Element?, Failure>) {
+      func emit(_ result: Result<Base.Element?, Error>) {
       let (resumptions, limitContinuation, demandContinuation, cancelled) = state.withLock {
         state -> ([Resumption], UnsafeContinuation<Bool, Never>?, UnsafeContinuation<Void, Never>?, Bool) in
         var resumptions = [Resumption]()
@@ -533,12 +531,12 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
 
     private func nextIteration(
       _ id: Int
-    ) async -> Result<AsyncShareSequence<Base>.Element?, AsyncShareSequence<Base>.Failure> {
+    ) async -> Result<Base.Element?, Error> {
       return await withTaskCancellationHandler {
         await withUnsafeContinuation { continuation in
           let (res, limitContinuation, demandContinuation, cancelled) = state.withLock {
             state -> (
-              Result<Element?, Failure>?, UnsafeContinuation<Bool, Never>?, UnsafeContinuation<Void, Never>?, Bool
+                Result<Base.Element?, Error>?, UnsafeContinuation<Bool, Never>?, UnsafeContinuation<Void, Never>?, Bool
             ) in
             guard let side = state.sides[id] else {
               return state.emit(.success(nil), limit: limit)
@@ -587,11 +585,11 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
           }
         }
       } catch {
-        emit(.failure(error as! Failure))
+        emit(.failure(error))
       }
     }
 
-    func next(isolation actor: isolated (any Actor)?, id: Int) async throws(Failure) -> Element? {
+    func next(isolation actor: isolated (any Actor)?, id: Int) async throws -> Base.Element? {
       let (factory, cancelled) = state.withLock { state -> ((@Sendable () -> sending Base.AsyncIterator)?, Bool) in
         switch state.iteratingTask {
         case .pending(let factory):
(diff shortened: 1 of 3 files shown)


stefanomondino:
@Akazm I think it's the same (I found it in the one you mentioned at the bottom of the readme, the author is the same).
I've mentioned the gist and not the entire library because it might be easier to integrate in a library like async-algorithms avoiding an explicit third party SPM dependency.

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -230,9 +228,9 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
       var generation = 0
       var sides = [Int: Side.State]()
       var iteratingTask: IteratingTask
-      private(set) var buffer = Deque<Element>()
+        private(set) var buffer = Deque<Base.Element>()
```
pyrtsa:
```suggestion
      private(set) var buffer = Deque<Base.Element>()
```

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -311,7 +309,7 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
       //   **Buffering Newest**: Appends if under the limit, otherwise removes the oldest and appends
       //
       // - Parameter element: The element to add to the buffer
-      mutating func enqueue(_ element: Element) {
+        mutating func enqueue(_ element: Base.Element) {
```
pyrtsa:
```suggestion
      mutating func enqueue(_ element: Base.Element) {
```

Sources/AsyncAlgorithms/AsyncShareSequence.swift
pyrtsa: Here's a suggested fix (for lines 592–605) which I believe works the same and does not seem to crash on any of the older iOS versions. For me, tests passed on iOS simulators 15.5, 16.4, 17.5, 18.6, and 26.0.

```suggestion
      let iteratingTask = state.withLock { state -> IteratingTask in
        defer {
          if case .pending = state.iteratingTask {
            state.iteratingTask = .starting
          }
        }
        return state.iteratingTask
      }
      if case .cancelled = iteratingTask { return nil }
      if case .pending(let factory) = iteratingTask {
```

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -478,15 +476,15 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
     }
 
     struct Resumption {
-      let continuation: UnsafeContinuation<Result<Element?, Failure>, Never>
-      let result: Result<Element?, Failure>
+        let continuation: UnsafeContinuation<Result<Base.Element?, Error>, Never>
+        let result: Result<Base.Element?, Error>
```
pyrtsa:
```suggestion
      let continuation: UnsafeContinuation<Result<Base.Element?, Error>, Never>
      let result: Result<Base.Element?, Error>
```

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -478,15 +476,15 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
...
 
       func resume() {
         continuation.resume(returning: result)
       }
     }
 
-    func emit(_ result: Result<Element?, Failure>) {
+      func emit(_ result: Result<Base.Element?, Error>) {
```
pyrtsa:
```suggestion
    func emit(_ result: Result<Base.Element?, Error>) {
```

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -533,12 +531,12 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
...
     private func nextIteration(
       _ id: Int
-    ) async -> Result<AsyncShareSequence<Base>.Element?, AsyncShareSequence<Base>.Failure> {
+    ) async -> Result<Base.Element?, Error> {
       return await withTaskCancellationHandler {
         await withUnsafeContinuation { continuation in
           let (res, limitContinuation, demandContinuation, cancelled) = state.withLock {
             state -> (
-              Result<Element?, Failure>?, UnsafeContinuation<Bool, Never>?, UnsafeContinuation<Void, Never>?, Bool
+                Result<Base.Element?, Error>?, UnsafeContinuation<Bool, Never>?, UnsafeContinuation<Void, Never>?, Bool
```
pyrtsa:
```suggestion
              Result<Base.Element?, Error>?, UnsafeContinuation<Bool, Never>?, UnsafeContinuation<Void, Never>?, Bool
```

Sources/AsyncAlgorithms/AsyncShareSequence.swift
stefanomondino: Thank you so much! I've included your changes in my branch and it seems like the crash is solved for me as well on ios 17.5

pyrtsa: Let's do this! 🙌

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -135,7 +133,7 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
     // - `continuation`: The continuation waiting for the next element (nil if not waiting)
     // - `position`: The consumer's current position in the shared buffer
     struct State {
-      var continuation: UnsafeContinuation<Result<Element?, Failure>, Never>?
+        var continuation: UnsafeContinuation<Result<Base.Element?, Error>, Never>?
```
pyrtsa:
```suggestion
      var continuation: UnsafeContinuation<Result<Base.Element?, Error>, Never>?
```

stefanomondino: @pyrtsa again? let me doublecheck what's going on on my end - I usually have swiftlint + swiftformat getting rid of those errors automatically

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -587,50 +585,49 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
...
-        #if swift(>=6.2)
-        if #available(macOS 26.0, iOS 26.0, tvOS 26.0, visionOS 26.0, *) {
-          task = Task(name: "Share Iteration") { [factory, self] in
-            await iterationLoop(factory: factory)
-          }
-        } else {
-          task = Task.detached(name: "Share Iteration") { [factory, self] in
-            await iterationLoop(factory: factory)
+    func next(isolation actor: isolated (any Actor)?, id: Int) async throws -> Base.Element? {
+        let iteratingTask = state.withLock { state -> IteratingTask in
```
pyrtsa: The body of this function needs some reindenting still, lines 593–630 are still indented too much.

Tests/AsyncAlgorithmsTests/TestShare.swift
```diff
@@ -94,13 +94,13 @@ final class TestShare: XCTestCase {
       gate1.open()
       await gate2.enter()
       // Consumer 1 reads first element
-      if let value = await iterator.next(isolation: nil) {
+      if let value = try await iterator.next(isolation: nil) {
         results1.withLock { $0.append(value) }
       }
       // Delay to allow consumer 2 to get ahead
-      try? await Task.sleep(for: .milliseconds(10))
+        try? await Task.sleep(nanoseconds: 10_000_000)
```
pyrtsa:
```suggestion
      try? await Task.sleep(nanoseconds: 10_000_000)
```

Sources/AsyncAlgorithms/AsyncShareSequence.swift
pyrtsa: PROTIP: Just copy the `.editorconfig` file I suggested in the earlier comment and place it at the root of the project, even if you didn't commit it. Xcode respects that while editing code.

Tests/AsyncAlgorithmsTests/TestShare.swift
```diff
@@ -572,6 +572,35 @@ final class TestShare: XCTestCase {
...
+      let shared: some AsyncSequence<Void, TestError> = FailingSequence(TestError.failure).share()
+      do {
+        for try await _ in shared {
+          XCTFail("Expected to not get here")
+        }
+      } catch {
+        XCTAssertEqual(error, TestError.failure)
+      }
+    } else {
+      _ = XCTSkip("This test is not available before 1.2")
```
pyrtsa: I think you need to throw it in order to mark the test as skipped:
```suggestion
      throw XCTSkip("This test is not available before 1.2")
```

Sources/AsyncAlgorithms/Shims.swift
```diff
@@ -0,0 +1,18 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
```
phausler:
```suggestion
// Copyright (c) 2025 Apple Inc. and the Swift project authors
```

phausler: There seems to be some other ones that I copy/pasta'd too .... oops...

Tests/AsyncAlgorithmsTests/Support/FailingSequence.swift
```diff
@@ -0,0 +1,26 @@
+//
+//  FailingSequence.swift
+//  swift-async-algorithms
+//
+//  Created by Stefano Mondino on 15/10/25.
+//
```
phausler, line 7: This should also be updated to the right attribution header:

```swift
//
// This source file is part of the Swift Async Algorithms open source project
//
// Copyright (c) 2025 Apple Inc. and the Swift project authors
// Licensed under Apache License v2.0 with Runtime Library Exception
//
// See https://swift.org/LICENSE.txt for license information
//
```

Tests/AsyncAlgorithmsTests/TestShare.swift
```diff
@@ -130,31 +130,31 @@ final class TestShare: XCTestCase {
...
     let consumer1 = Task {
       var iterator = shared.makeAsyncIterator()
       gate2.open()
       await gate1.enter()
-      while let value = await iterator.next(isolation: nil) {
+      while let value = await iterator.next() {
         results1.withLock { $0.append(value) }
         // Add some delay to consumer 1
-        try? await Task.sleep(for: .milliseconds(1))
+        try? await Task.sleep(nanoseconds: 1_000_000)
```
phausler: It might be worthwhile to keep the test suite requiring a higher version for ease of deployment (and validating the newer versions folks will in time all be using) and it avoids needing to use crusty old versions of the sleep method

Tests/AsyncAlgorithmsTests/TestShare.swift
```diff
@@ -572,6 +572,35 @@ final class TestShare: XCTestCase {
...
+      let shared: some AsyncSequence<Void, TestError> = FailingSequence(TestError.failure).share()
+      do {
+        for try await _ in shared {
+          XCTFail("Expected to not get here")
+        }
+      } catch {
+        XCTAssertEqual(error, TestError.failure)
+      }
+    } else {
+      throw XCTSkip("This test is not available before 1.2")
```
phausler: I don't think this can actually hit right?

Sources/AsyncAlgorithms/AsyncShareSequence.swift
```diff
@@ -130,12 +129,16 @@ where Base.Element: Sendable, Base: SendableMetatype, Base.AsyncIterator: Sendab
   //   **Usage**: Tracks buffer position and manages async continuations
   //   **Cleanup**: Automatically unregisters and cancels pending operations on deinit
   final class Side {
+    // Due to a runtime crash in 1.0 compatible versions, it's not possible to handle
+    // a generic failure constrained to Base.Failure. We handle inner failure with a `any Error`
```
phausler:
```suggestion
    // a generic failure constrained to Base.Failure. We handle inner failure with a `any Error`
```

Tests/AsyncAlgorithmsTests/TestShare.swift
pyrtsa: It certainly can! The overall test suite of this file now passes tests on OS versions as old as iOS 14; it's just this single test case exercising typed throws which really requires iOS 18 or similar, and skipped otherwise.

stefanomondino: The issue is that XCTest does not support availability macro directly on the function due to how tests are automatically "expanded" from the suite. This means that running this test on older platforms would result in an empty one. 
It felt more appropriate to expose it as a skipped test, but it's not a big deal to remove it

pyrtsa: (Just a style comment, I think this is cleaner if expressed as `guard #available(...) else { throw ... }` at the start of the test function.)

Tests/AsyncAlgorithmsTests/TestShare.swift
stefanomondino: @phausler I think it depends on how the general workflow for testing is set up. I think it's useful being able to run tests on legacy platforms if we support them because we found critical issues and crashes there.

I've set AsyncAlgorithms 1.1 to be compatible with iOS 14.0 because - thanks to the tests being able to run there - I discovered issues on iOS 13.0. 
I'll create a backportable local extension on `Task` to avoid the nanoseconds initializer.

pyrtsa:
> can you please doublecheck my last commits? Do you see any possible issue? I was thinking there's no need for this operator to be available only on latest compiler versions

LGTM! 🙌

alessionossa:
Any update on this? [AsyncAlgorithms 1.1.1](https://github.com/apple/swift-async-algorithms/releases/tag/1.1.1) has been released with support for iOS 18+ only.

phausler:
So re-reading through this: as it currently stands it is not mergable since this is breaking. I think there are ways to modify what is happening here to make it not breaking but currently I don't have really much allotted time to devote to refining that.

This type of failure is a no-go type thing:
```
API breakage: func AsyncSequence.share(bufferingPolicy:) has return type change from some Swift.Sendable & _Concurrency.AsyncSequence<Self.Element, Self.Failure> to AsyncAlgorithms.AsyncShareSequence<Self>
```

At this point it might be worth while to do a few things (perhaps as a starting point to move towards a mergable version). 1) focus just on one API (e.g. just share) for now 2) introduce it as replication and invert the guard for the API (so basically copy/paste the entire thing focused to pre swift-6.2 requirements and gate that accordingly).

stefanomondino:
@phausler  I get the point (in general) about the error - the API is returning a different type. And probably is a matter of ABI stability or something like that - I don't have enough knowledge about that 

However `AsyncAlgorithms.AsyncShareSequence<Self>` is the exact same type that was returned before without being "hidden" by `some AsyncSequence<Element, Failure> & Sendable` (and that's why it's now public). So - with my previous disclaimer in mind - it's not exactly clear to me how this could be a "real" breaking change for already existing APIs.
It's not like changing a return value from `String` to `Int` or something like that - consumers were unaware of the true inner nature of the returned sequence before, they can keep ignoring it without even noticing. 

That being said, I've put together an example of what can be done following your suggestions. It needs more documentation to let people better understand the differences, but before that I'd rather have a feedback. 

Please note again that I had to change the availability macros, using 1.1 for this new operator and 1.2 for everything else. 1.0 cannot be used here due to some limitations on iOS 13.0. 

@pyrtsa let me also know what you think please.

Issue #381, opened 2025-11-29
Proposal: Add flatMapLatest operator


peterfriese:
I would like to propose adding a flatMapLatest operator to AsyncSequence.

## Motivation
When transforming elements of an asynchronous sequence into new asynchronous sequences, there are cases where only the results from the most recent transformation are relevant. A common example is a search-as-you-type interface where each keystroke triggers a network request. If a new request starts, the previous one becomes irrelevant and should be cancelled to save resources and avoid displaying stale data.

## Proposed Solution
Add a flatMapLatest operator that transforms each element from the base AsyncSequence into a new inner AsyncSequence. When a new element is produced by the base sequence, iteration on the current inner sequence is cancelled, and iteration begins on the newly created sequence.

## Example

```
let searchResults = searchQueries.flatMapLatest { query in
  searchAPI(query)
}

for await result in searchResults {
  displayResults(result) // Only latest results displayed
}
```

I've prepared an Evolution proposal and a prototype implementation including tests (see https://github.com/peterfriese/swift-async-algorithms/tree/flatMapLatest/statemachine)

pyrtsa:
The PR was merged. Why is this issue still open?

peterfriese:
Closed via #382

PR #382, opened 2025-11-29
feat: Add AsyncSequence.flatMapLatest operator

peterfriese:
This PR introduces the `flatMapLatest` operator for `AsyncSequence`, refined through collaboration with Philippe Hausler.

## What's included:

*   **`flatMapLatest` operator**: A new operator that transforms elements of an asynchronous sequence into new asynchronous sequences, automatically cancelling previous inner sequences when a new element arrives from the outer sequence.
*   **Refined API Surface** (Credit: Philippe Hausler):
    *   Uses **opaque return types** (`some AsyncSequence`) for better encapsulation.
    *   Full support for **typed throws**, ensuring type-safe error propagation.
    *   Added **`switchToLatest()`** shorthand for `AsyncSequence` of `AsyncSequence`.
*   **Performance & Robustness**:
    *   **High-performance State Machine**: Implementation uses a state machine pattern to ensure thread-safe operation and efficient coordination.
    *   **Race Condition Fix** (Credit: Peter Friese): Resolved a critical concurrency issue where inner stream producers could yield values after cancellation, causing potential crashes or inconsistent state.
*   **Comprehensive Testing**:
    *   Extensive unit tests covering error propagation, cancellation, and empty sequences.
    *   Added a specific **concurrency regression test** (`test_concurrency_crash_repro`) to ensure the race condition fix remains effective.
*   **Documentation & Evolution**:
    *   Updated **Evolution Proposal** (`Evolution/00nn-flatMapLatest.md`) reflecting the pitchable state with the refined API.
    *   Added **`mapError`** operator (dependency) to support the refined implementation.

## Motivation:

The `flatMapLatest` operator (and its shorthand `switchToLatest`) addresses the common problem of managing asynchronous operations where only the result of the most recent operation is relevant. This is essential for building responsive, data-driven user interfaces in Swift.

This PR is an implementation of #381

---
*This PR represents an integrated effort combining the original proposal with refinements for modern Swift Concurrency patterns.*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ DerivedData/
 .swiftpm
 Package.resolved
 .swift-version
+test_output.log


Sources/AsyncAlgorithms/FlatMapLatest/AsyncFlatMapLatestSequence.swift
```diff
@@ -0,0 +1,55 @@
...
+  }
+  
+  public struct Iterator: AsyncIteratorProtocol, Sendable {
+    let storage: FlatMapLatestStorage<Base, Inner>
+    
+    init(base: Base, transform: @escaping @Sendable (Base.Element) -> Inner) {
+      self.storage = FlatMapLatestStorage(base: base, transform: transform)
+    }
+    
+    public func next() async throws -> Element? {
```
phausler: couple of notes here that might inform the rest of the potential pitch:
the next method should likely be the typed throws variant (because as it currently stands this algorithm always throws, which is less than ideal if the base components never throw)
that would then make the overload for the old variant of next (the unaddorned non-typed throws version as you have written) would be `rethrows` not `throws`

Sources/AsyncAlgorithms/FlatMapLatest/FlatMapLatestStateMachine.swift
```diff
@@ -0,0 +1,439 @@
...
+  enum NextAction {
+    case returnElement(Element)
+    case returnNil
+    case throwError(Error)
+    case startOuterTask(Base)
+    case suspend
+  }
+  
+  enum Action {
+    case startInnerTask(Inner, generation: Int, previousTask: Task<Void, Never>?, previousContinuation: UnsafeContinuation<Void, Error>?)
```
phausler: I would like to potentially have the flatMap and switchToLatest families eventually only be driven by one task, do you think that is something we could transition to from this implementation?

peterfriese: I tried following the implementation of the other algorithms.

Maybe this is something we can look into once the initial version has been merged?

Sources/AsyncAlgorithms/FlatMapLatest/AsyncFlatMapLatestSequence.swift
peterfriese: +1 on using typed throws, thanks for the suggestion! I see you implemented this in PR https://github.com/apple/swift-async-algorithms/pull/395, so I will not change it here for the time being.

phausler: so my pr was just to update things, feel free to merge my branch into yours and continue iterating. the major concerns here are more for the review with regards to the closure throwing or being async etc. I can help out with getting the implementation over the line.

Sources/AsyncAlgorithms/FlatMapLatest/FlatMapLatestStateMachine.swift
phausler: yes this could be an optimization later down the road.

Sources/AsyncAlgorithms/FlatMapLatest/AsyncFlatMapLatestSequence.swift
peterfriese: I cherry-picked your suggestions into this PR, thanks!

PR #384, opened 2025-12-09
Make MPSCAsyncChannel `#isolation` methods use nonisolated(nonsending)`.

gottesmm:
We can perform more aggressive optimizer optimizations with nonisolated(nonsending) than #isolation. So use that instead to get some easy perf wins.

### DIFF
--- a/Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannel+Internal.swift
+++ b/Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannel+Internal.swift
@@ -391,7 +391,7 @@ extension MultiProducerSingleConsumerAsyncChannel {
     }
 
     @inlinable
-    func next(isolation: isolated (any Actor)? = #isolation) async throws -> Element? {
+    nonisolated(nonsending) func next() async throws -> Element? {
       let action = self._stateMachine.withLock {
         $0.next()
       }
@@ -431,7 +431,7 @@ extension MultiProducerSingleConsumerAsyncChannel {
     }
 
     @inlinable
-    func suspendNext(isolation: isolated (any Actor)? = #isolation) async throws -> Element? {
+    nonisolated(nonsending) func suspendNext() async throws -> Element? {
       try await withTaskCancellationHandler {
         try await withUnsafeThrowingContinuation { (continuation: UnsafeContinuation<Element?, Error>) in
           let action = self._stateMachine.withLock {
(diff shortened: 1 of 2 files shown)


Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannel.swift
```diff
@@ -710,9 +708,7 @@ extension MultiProducerSingleConsumerAsyncChannel.ChannelAsyncSequence where Ele
     }
 
     @inlinable
-    mutating func next(
```
FranzBusch, line 713: I don't think we can change this one since this is a protocol requirement of `AsyncSequence` unless `nonisolated(nonsending)` methods are capable of fulfilling `isolated` protocol requirement

Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannel+Internal.swift
```diff
@@ -431,7 +431,7 @@ extension MultiProducerSingleConsumerAsyncChannel {
     }
 
     @inlinable
-    func suspendNext(isolation: isolated (any Actor)? = #isolation) async throws -> Element? {
+    nonisolated(nonsending) func suspendNext() async throws -> Element? {
```
jamieQ, line 434: in my recent testing, it seems `nonisolated(nonsending)` does not grant the ability to actually procure a reference to the `isolated` parameter in a way that allows isolating closures in the same manner that capturing an explicit `isolated` parameter does. i suppose in this case it doesn't matter since the parameter isn't being captured by the body closures, but that does make me wonder if those closures are actually not being isolated in the intended manner currently.

FranzBusch: Could you briefly expand what closures you are worried are not isolated in particular?

jamieQ: so, the concern would be that even though the method inherits isolation by default (in the original formulation), the isolated parameter is unused within the body closures. this is true in the prior version with an isolated parameter, and seems to also be true even after converting to implicit caller isolation inheritance. this means the actual 'work' invoked by the `with*` runtime methods isn't isolated, so IIUC it's going to require an executor hop if the caller is actor-isolated. this can be seen in this sort of example: https://swift.godbolt.org/z/GP4dPhfKv.

FranzBusch: Ah so you are talking about the `withTaskCancellationHandler` and `withCheckedContinuation` handler. Both of them have open PRs to address these issues https://github.com/swiftlang/swift/pull/85191 & https://github.com/swiftlang/swift/pull/85518

Edit: I see that you left a comment on the latter PR. The hop should be eliminated and I think @gottesmm is looking into that.

Issue #356, opened 2025-07-09
FIFO not respected in `AsyncChannel`


alessionossa:
While testing `AsyncChannel` behavior to determine if it was a good match for my use case, I noticed that when `AsyncChannel.send(_:)` is called from different `Task`s, the order of the output is inconsistent and unpredictable, even if the schedule happens on a specific actor and so does the reading.

The following script sends 500 elements, one after the other, each in its own `Task`. This recreates a common scenario of the real word where new elements generates from user inputs in a synchronous context and the call to `send(_:)` has to be wrapped in a Task to bridge sync and async worlds. This behavior is also described at [[Pitch] Async buffered channel - Related Projects / Swift Async Algorithms - Swift Forums](https://forums.swift.org/t/pitch-async-buffered-channel/59854).

```swift
import Foundation
import AsyncAlgorithms

@globalActor actor ConsumerActor: GlobalActor {
    static let shared = ConsumerActor()
}

func testAsyncChannel() async throws {
    let channel = AsyncChannel<Int>()

    let task1 = Task { @ConsumerActor in        
        var count = 0
        
        for await value in channel {
            count += 1
            if value != count {
                print(value, "on ConsumerActor does not match", count)
            }
            // print("AsyncChannel value: ", value)
        }
    }
    
    for element in 1...500 {
        Task { @MainActor in
            // print("Sending \(element)")
            await channel.send(element)
        }
        // try await Task.sleep(for: .nanoseconds(30000)) // For values above 30000 ns, the order is respected.
    }
}

try await testAsyncChannel()

try await Task.sleep(for: .seconds(2))
```

The output this script is:
```plain
3 on ConsumerActor does not match 1
4 on ConsumerActor does not match 2
5 on ConsumerActor does not match 3
6 on ConsumerActor does not match 4
7 on ConsumerActor does not match 5
1 on ConsumerActor does not match 6
8 on ConsumerActor does not match 7
12 on ConsumerActor does not match 8
13 on ConsumerActor does not match 9
[...]
```
Showing that the order the elements are emitted by the `AsyncChannel` does not correspond to the ordinal numbers.

The expectation is that the elements emitted by the `AsyncChannel` are in the same order `AsyncChannel.send(_:)` has been called. This happens when calls to `AsyncChannel.send(_:)` are at least 30000 ns.

[...]

PR #388, opened 2025-12-15
[WIP]: inherit caller isolation in some AsyncChannel methods

jamieQ:
This change adopts caller isolation inheritance via parameter isolation in Async{Throwing}Channel and corresponding AsyncIterator types.

Resolves: https://github.com/apple/swift-async-algorithms/issues/356

### DIFF
--- a/Sources/AsyncAlgorithms/Channels/AsyncChannel.swift
+++ b/Sources/AsyncAlgorithms/Channels/AsyncChannel.swift
@@ -35,7 +35,10 @@ public final class AsyncChannel<Element: Sendable>: AsyncSequence, Sendable {
   /// If the channel is already finished then this returns immediately.
   /// If the task is cancelled, this function will resume without sending the element.
   /// Other sending operations from other tasks will remain active.
-  public func send(_ element: Element) async {
+  public func send(
+    isolation: isolated (any Actor)? = #isolation,
+    _ element: Element
+  ) async {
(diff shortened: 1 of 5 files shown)


jamieQ:
cc @FranzBusch @phausler – this is largely a PoC of how we might address some executor hops that could lead to unexpected event orderings. i'd like to hear your thoughts on the approach. in particular:

1. are there concerns with adding defaulted isolated parameters into existing API?
2. is tackling this sort of thing in a piecemeal fashion palatable, or have you considered a wholesale audit of the repo to address the class of issue?
    - relatedly, do you have thoughts regarding opting into `NonisolatedNonsendingByDefault` at some point (though i'm personally not sure it's totally ready for all use cases yet)?

jamieQ:
> Did you give it a try to compile this repo with that default enabled and see if we run into any issues?

@FranzBusch i did a very brief test, and there were many errors involving non-sendability that immediately surfaced for isolation boundary closures (i believe possibly fixed in https://github.com/swiftlang/swift/pull/85603 and friends). also, basically every async iterator `next()` implementation errors b/c the existing implementations use the non-isolation-inheriting form.

IIUC just enabling that feature now also would not resolve the issues with hops when calling the concurrency runtime functions (`with*(...)`) until they are somehow changed. it's not clear to me if any of the in-flight PRs to do that will ensure the semantic change is realized on older runtimes though.

PR #389, opened 2025-12-17
Support for Wasm using Swift SDKs that don't support pthreads

scottmarchant:
# Summary

This PR adds support for compiling `swift-async-algorithms` to wasm using the [Swift SDK for WebAssembly](https://www.swift.org/documentation/articles/wasm-getting-started.html) - but without pthreads and specialized command such as `swift build -Xcc -D_WASI_EMULATED_PTHREAD`.

This is achieved by eliding mutex locking for single-thread non-pthread WASI platforms, similar to the approach used in [swift-log](https://github.com/apple/swift-log/blob/main/Sources/Logging/Locks.swift#L30). This enables building `swift-async-algorithms` with the default Swift for WebAssembly configurations

This PR is [part of a larger effort](https://github.com/PassiveLogic/swift-web-examples/issues/1) by a company called PassiveLogic to enable broad support for Swift WebAssembly compilation.

# Details

- Added additional compilation conditionals to elide mutex locking for WASI platforms that lack pthread support and carry a decently strong guarantee of single-threaded operation.
- Removed specialized wasm build command customization from CI configuration, since `swift-async-algorithms` will now compile with the vanilla default configuration with these changes.

# Testing done

- [x] Cleaned up swiftformat lint on modified lines of change.
- [x] Verified unit tests still pass with these changes
- [x] Verified `swift build` completes without errors
- [x] Verified `swift build --swift-sdk wasm32-unknown-wasip1 --target AsyncAlgorithms` completes without errors
- [x] Verified `swift build --swift-sdk wasm32-unknown-wasip1-threads --target AsyncAlgorithms` completes without errors
- [x] Verified CI passes using [CI check on separate draft PR](https://github.com/PassiveLogic/swift-async-algorithms/actions/runs/20287929178/job/58265876675?pr=1)

# Impact Risk

There should be no negeative impact risk for existing targets. The changes only affect compilation for WASI platforms that don't support pthread, which was previously unsupported.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -15,7 +15,6 @@ jobs:
       linux_exclude_swift_versions: "[{\"swift_version\": \"5.9\"}, {\"swift_version\": \"5.10\"}]]"
       windows_exclude_swift_versions: "[{\"swift_version\": \"5.9\"}]"
       enable_wasm_sdk_build: true
-      wasm_sdk_build_command: swift build -Xcc -D_WASI_EMULATED_PTHREAD
 
   soundness:
     name: Soundness
--- a/Sources/AsyncAlgorithms/Locking.swift
+++ b/Sources/AsyncAlgorithms/Locking.swift
@@ -21,6 +21,9 @@ import WinSDK
 import Bionic
 #elseif canImport(wasi_pthread)
 import wasi_pthread
+#elseif canImport(WASILibc)
+// No locking on WASILibc provided by the current Swift for WebAssembly SDK as of Dec 2025.
+// That SDK is also single-threaded, so we can safely avoid locking altogether.
 #else
 #error("Unsupported platform")
 #endif
@@ -30,6 +33,8 @@ internal struct Lock {
   typealias Primitive = os_unfair_lock
   #elseif canImport(Glibc) || canImport(Musl) || canImport(Bionic) || canImport(wasi_pthread)
   typealias Primitive = pthread_mutex_t
+  #elseif canImport(WASILibc)
+  // This WASILibc variation is single threaded, provides no locks
   #elseif canImport(WinSDK)
   typealias Primitive = SRWLOCK
   #else
@@ -49,6 +54,8 @@ internal struct Lock {
     #elseif canImport(Glibc) || canImport(Musl) || canImport(Bionic) || canImport(wasi_pthread)
     let result = pthread_mutex_init(platformLock, nil)
     precondition(result == 0, "pthread_mutex_init failed")
+    #elseif canImport(WASILibc)
+    // This WASILibc variation is single threaded, provides no locks
     #elseif canImport(WinSDK)
     InitializeSRWLock(platformLock)
     #else
@@ -69,6 +76,9 @@ internal struct Lock {
     os_unfair_lock_lock(platformLock)
     #elseif canImport(Glibc) || canImport(Musl) || canImport(Bionic) || canImport(wasi_pthread)
     pthread_mutex_lock(platformLock)
+    #elseif canImport(WASILibc)
+    // This WASILibc variation is single threaded, provides no locks
+    return
     #elseif canImport(WinSDK)
     AcquireSRWLockExclusive(platformLock)
     #else
@@ -82,6 +92,9 @@ internal struct Lock {
     #elseif canImport(Glibc) || canImport(Musl) || canImport(Bionic) || canImport(wasi_pthread)
     let result = pthread_mutex_unlock(platformLock)
     precondition(result == 0, "pthread_mutex_unlock failed")
+    #elseif canImport(WASILibc)
+    // This WASILibc variation is single threaded, provides no locks
+    return
     #elseif canImport(WinSDK)
     ReleaseSRWLockExclusive(platformLock)
     #else


.github/workflows/pull_request.yml
```diff
@@ -15,7 +15,6 @@ jobs:
       linux_exclude_swift_versions: "[{\"swift_version\": \"5.9\"}, {\"swift_version\": \"5.10\"}]]"
       windows_exclude_swift_versions: "[{\"swift_version\": \"5.9\"}]"
       enable_wasm_sdk_build: true
-      wasm_sdk_build_command: swift build -Xcc -D_WASI_EMULATED_PTHREAD
```
scottmarchant, line 18: @MaxDesiatov Just wanted to call this to your attention. With the changes in this PR, pthreads are no longer required to compile `swift-async-algorithms`. But I verified the pthread approach still gets compiled in when available.

Remove this build command customization seemed like the right thing to do with these changes, but interested to know your thoughts here.

Sources/AsyncAlgorithms/Locking.swift
```diff
@@ -19,9 +19,15 @@ import Musl
 import WinSDK
 #elseif canImport(Bionic)
 import Bionic
-#elseif canImport(wasi_pthread)
+#elseif canImport(WASILibc)
+#if canImport(wasi_pthread)
+import WASILibc
 import wasi_pthread
 #else
+// No locking on WASILibc provided by the current Swift for WebAssembly SDK as of Dec 2025.
```
scottmarchant, line 25: Note that we're using a similar approach in other [swift repos](https://github.com/apple/swift-log/blob/main/Sources/Logging/Locks.swift#L30)

.github/workflows/pull_request.yml
MaxDesiatov: @kateinoigakukun WDYT? IIRC `import wasi_pthread` is available in recent single-threaded Swift SDKs too and these conditional imports are no longer needed?

kateinoigakukun: I'd say it's better to rely on the pthread stub API for better consistency with other platforms. If we don't need to support Swift 6.2, we can unconditionally import `wasi_pthread` regardless of `p1` or `p1-threads`.

Sources/AsyncAlgorithms/Locking.swift
kateinoigakukun: We need to use pthread's mutex API based implementation when `wasi_pthread` is available.

.github/workflows/pull_request.yml
scottmarchant: @MaxDesiatov @kateinoigakukun Would it make sense to update this build command in `swiftlang/github-workflows` when we're able to unconditionally import `wasi_pthread`, and remove the customization here? That way, the build command here will remain consistent with the other wasm CI builds.

Happy to revert this, just wanted to mention that thought before reverting this line. I believe this repo will compile fine in CI either way as long as the remaining changes are in place.

Sources/AsyncAlgorithms/Locking.swift
scottmarchant: Ah right, I should update this import pattern to match the patterns below. It will be a lot more readable that way as well. Thanks @kateinoigakukun!

.github/workflows/pull_request.yml
MaxDesiatov: If that option doesn't belong here in this repository, but is required for `import wasi_pthread` to work (I haven't confirmed this myself), then IMO it should be in the Swift Driver codebase, or a Swift SDK toolset (generated by Swift SDK Generator), or somewhere on that level so that everyone gets that behavior by default.

scottmarchant: @MaxDesiatov +1 for enabling the emulated pthread behavior by default. I think it would help a lot more code out there compile.

Sources/AsyncAlgorithms/Locking.swift
scottmarchant: @kateinoigakukun I adjusted this to simplify the import and align with your comment.

.github/workflows/pull_request.yml
scottmarchant: I'm leaving this as is for now. But if either of you want me to revert, please let me know. Happy to change this.

I confirmed using sdk `swift-6.3-DEVELOPMENT-SNAPSHOT-2025-12-07-a_wasm` that the command `swift build -Xcc -D_WASI_EMULATED_PTHREAD --swift-sdk wasm32-unknown-wasip1 --target AsyncAlgorithms` will make `canImport(wasi_pthread)` truthy, whereas `swift build --swift-sdk wasm32-unknown-wasip1 --target AsyncAlgorithms` makes `canImport(wasi_pthread)` false.

phausler:
@MaxDesiatov I can revert and/or land a fix.

Sources/AsyncAlgorithms/Locking.swift
```diff
@@ -30,6 +33,8 @@ internal struct Lock {
   typealias Primitive = os_unfair_lock
   #elseif canImport(Glibc) || canImport(Musl) || canImport(Bionic) || canImport(wasi_pthread)
   typealias Primitive = pthread_mutex_t
+  #elseif canImport(WASILibc)
+  // This WASILibc variation is single threaded, provides no locks
```
scottmarchant, line 37: @MaxDesiatov I think actually this comment might be mostly correct, because it is an `#elseif` in the case where `canImport(wasi_pthread)` is false. Took me a minute to think about this, since it is so confusingly similar to the older [swift-log pattern](https://github.com/apple/swift-log/blob/main/Sources/Logging/Locks.swift#L30) we're [updating](https://github.com/apple/swift-log/pull/398).

Would you like me to create a PR to align the patterns more closely?

I think this pattern would actually stay the same, but the comment should probably read:

```
// Here, we can not import wasi_pthread, but we can import WASILibc, so
// this is a variation of WASILibc without pthread support. As such, the
// target does not have locks nor mechanisms to become multi-threaded,
// so no locks are needed.
```

Thoughts on this?

scottmarchant: Another option is to move the `canImport(wasi_pthread)` from the line above into a nested wasi condition, and then we would match the swift-nio pattern more closely:

```
…
#elseif canImport(WASILibc)
#if canImport(wasi_pthread)
import wasi_pthread
#else
// This WASILibc variation does not have pthread support, provides no locks, and
// is assumed to be single threaded.
#endif
#else
…
```

MaxDesiatov: Yes, either would be fine if you have a moment to update it. Thank you so much!

scottmarchant: Sounds good, I'll submit a PR shortly with the first option that updates the code comments. Not sure I want to risk changing the imports in this file at the moment.

scottmarchant: @MaxDesiatov @phausler Updated the comments in a follow on PR: https://github.com/apple/swift-async-algorithms/pull/393

PR #390, opened 2025-12-26
Implement next(isolation:) in AsyncAdjacentPairsSequence

alex-waldron:
The next(isolation:) method allows the iterator to inherit the caller's isolation context, avoiding unnecessary isolation hops.


Tests/AsyncAlgorithmsTests/TestAdjacentPairs.swift
```diff
@@ -11,6 +11,7 @@
 
 import XCTest
 import AsyncAlgorithms
+import Observation
```
phausler: I think the test does not really need `Observations` to work - we have other things that dont have this added dependency.

alex-waldron: Nope it doesn't. Its just what I was using when I discovered the issue. I replaced it with an existing test async sequence in [425dba6](https://github.com/apple/swift-async-algorithms/pull/390/commits/425dba6fb2fdfc878db7f9efa6130e62f17dd4c1)

phausler:
@swift-ci please test

PR #396, opened 2026-01-29
Introduce a `withDeadline` method

FranzBusch:
## Motivation

Asynchronous operations in Swift can run indefinitely, which creates several problems in real-world applications. Currently, developers must implement timeout logic manually using task groups and clock sleep operations, resulting in verbose, error-prone code that's difficult to compose with surrounding async contexts. Each implementation must carefully handle cancellation, error propagation, and race conditions between the operation and timer.

## Modifications

This PR introduces a `withTimeout` method.

## Result

An ecosystem wide timeout solution for asynchronous operations.


Sources/AsyncAlgorithms/Timeout/Timeout.swift
```diff
@@ -0,0 +1,142 @@
...
+  ) { group in
+    let body = body.takeSending()!
+    group.addTask {
+      do {
+        return .success(Disconnected(value: try await body()))
+      } catch let error as Failure {
+        return .error(error)
+      } catch {
+        // The compiler doesn't understand that body only throws Failure
+        fatalError()
```
CodaFi: It can't infer it but you can certainly annotate it and let type _checking_ help you

```
      do throws(Failure) {
        return .success(Disconnected(value: try await body()))
      } catch let error {
        return .error(error)
      }
```

FranzBusch: Thanks. I always forget that you can put a `throws()` after `do`. I saw it somewhere before and knew there was a way to help the compiler but couldn't figure it out.

Sources/AsyncAlgorithms/Timeout/Timeout.swift
```diff
@@ -0,0 +1,179 @@
...
+    group.addTask {
+      do throws(Failure) {
+        return .success(Disconnected(value: try await body()))
+      } catch {
+        return .error(error)
+      }
+    }
+    group.addTask {
+      do {
+        try await clock.sleep(for: timeout, tolerance: tolerance)
```
ph1ps: I think we've talked about this in the other PR already, but I don't remember the outcome: would it make sense to also add the `withTimeout` overload that takes an Instant rather than the Duration? This timeout comes with the "impreciseness" of adding the little "spinning off child task" delay on top of the timout Duration.

Sources/AsyncAlgorithms/Timeout/Timeout.swift
```diff
@@ -0,0 +1,179 @@
...
+///
+/// - Returns: The result of the operation if it completes before the timeout expires.
+///
+/// - Throws: A ``TimeoutError`` containing the underlying error if the operation throws or times out.
+#if compiler(>=6.3)
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, *)
+nonisolated(nonsending) public func withTimeout<Return, Failure: Error, Clock: _Concurrency.Clock>(
+  in timeout: Clock.Instant.Duration,
+  tolerance: Clock.Instant.Duration? = nil,
+  clock: Clock = .continuous,
```
ph1ps: I think the `Duration` variant of `withTimeout` though could take `any Clock<Duration>` rather than some.

Imagine this model:

```
class Model {

  func doSomething() async throws {
    try await withTimeout(in: .seconds(5)) { ... }
  }
}
```

If you were to unit test `doSomething`, you would normally try to inject a custom clock, to not make the unit test wait for 5 seconds.

Either with some sort of dependency injection, or with simply passing it along the initializer you could do this:

```
class Model {

  let clock: any Clock<Duration>

  func doSomething() async throws {
    try await withTimeout(in: .seconds(5), clock: clock) { ... }
  }
}
```

At call site:
```
let model1 = Model(clock: ContinuousClock())
let model2 = Model(clock: TestClock())
```

Sources/AsyncAlgorithms/Timeout/Timeout.swift
FranzBusch: Yeah this was raised in the forums thread again. Having a `withDeadline` method.

Sources/AsyncAlgorithms/Timeout/Timeout.swift
FranzBusch: I think you can do this already with the caveat that `Duration` cannot be used on an arbitrary clock so you have to constrain it. This works with the existing code:

```swift
@Test(arguments: [ContinuousClock()])
func dependencyInjection(clock: any Clock<Duration>) async throws {
  try await self.concreteClock(clock: clock)
}

private func concreteClock(clock: some Clock<Duration>) async throws {
  try await withTimeout(
    in: .seconds(10),
    clock: clock
  ) {
    try await Task.sleep(for: .milliseconds(10))
  }
}
```

Sources/AsyncAlgorithms/Timeout/Disconnected.swift
```diff
@@ -0,0 +1,27 @@
...
+  private nonisolated(unsafe) var value: Value?
+
+  init(value: consuming sending Value) {
+    self.value = .some(value)
+  }
+
+  consuming func take() -> sending Value {
+    self.value.takeSending()!
+  }
+}
```
sliemeobn: are we sure this is safe? 
since both `Disconnected` and `Value` are copyable types, I don't see what would prevent copies of this ending up in different isolations.

FranzBusch: That's right. This type would ideally need to be `~Copyable` for this to be entirely safe. Sadly `withTaskGroup` doesn't support `~Copyable` types right now. A way around this would be to box the `Disconnected` but that incurs an allocation. I wanna pitch the non-copyable `Disconnected` type in a follow up pitch to this package.

sliemeobn: maybe you could add a few "do not try this at home" comments to make sure this does not spread into use cases where bad thing will happen.

Sources/AsyncAlgorithms/Timeout/Timeout.swift
```diff
@@ -0,0 +1,179 @@
...
+  }
+}
+#elseif compiler(>=6.2)
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, *)
+nonisolated(nonsending) public func withTimeout<Return, Failure: Error, Clock: _Concurrency.Clock>(
+  in timeout: Clock.Instant.Duration,
+  tolerance: Clock.Instant.Duration? = nil,
+  clock: Clock,
+  body: nonisolated(nonsending) () async throws(Failure) -> Return
+) async throws(TimeoutError<Failure>) -> Return {
```
nonameplum: On Swift `Swift version 6.2.3 (swift-6.2.3-RELEASE)`, I get a compilation error:

```
Cannot convert '@Sendable () async throws(Failure) -> Return' to 'nonisolated(nonsending) @Sendable () async throws(Failure) -> Return' because crossing of an isolation boundary requires parameter and result types to conform to 'Sendable' protocol
```

Do I miss something?

Sources/AsyncAlgorithms/Deadline/Deadline.swift
```diff
@@ -0,0 +1,314 @@
...
+    /// This case indicates the operation failed on its own without the deadline
+    /// being reached.
+    case operationFailed(OperationError)
+
+    public var description: String {
+      switch self {
+      case .deadlineExceeded(let error):
+        return "\(error)"
+      case .operationFailed(let error):
+        return "\(error)"
```
benlings, line 58: Should the `description` indicate which case was the cause?

FranzBusch:
Closing in favor of the stdlib proposal

PR #399, opened 2026-02-10
Fix a data race error with the internal `Optional.takeSending`

FranzBusch:
We added a little helper to move a non-Sendable type into an optional and take it out as sending so that it is disconnected. This was actually a data race hole which got fixed by the compiler so we need to use a proper disconnected abstractions.

### DIFF
--- a/Sources/AsyncAlgorithms/Internal/Disconnected.swift
+++ b/Sources/AsyncAlgorithms/Internal/Disconnected.swift
@@ -1,0 +1,38 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=6.1)
+// This is a helper type to move a non-Sendable value across isolation regions.
+@usableFromInline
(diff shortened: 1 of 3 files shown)


Sources/AsyncAlgorithms/Internal/Disconnected.swift
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=6.1)
+// This is a helper type to move a non-Sendable value across isolation regions.
+@usableFromInline
+struct Disconnected<Value: ~Copyable>: ~Copyable, Sendable {
```
phausler, line 15: eventually we should smush this together with some of the other boxes we have into one type for it...

Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannel+Internal.swift
```diff
@@ -1741,7 +1732,8 @@ struct SendableConsumeOnceBox<Wrapped> {
 
   @inlinable
   consuming func take() -> sending Wrapped {
-    return self.wrapped.takeSending()!
+    nonisolated(unsafe) let value = self.wrapped.take()!
```
phausler, line 1735: I wish there was a better way to spell this

phausler:
yes, I was planning on doing that today; sadly yesterday I was rather swamped with other tasks.

PR #400, opened 2026-02-13
Add new experimental `_AsyncStreaming` module

FranzBusch:
This PR adds an experimental module for asynchronous streaming interfaces as a potential evolution to async sequence. These APIs are considered highly unstable and are mostly for experimentation purposes; hence, they are part of an underscored module and all APIs are hidden behind a trait. I expect significant changes to those APIs to offer more functionality such as gather scatter and more.

### DIFF
--- a/Sources/AsyncStreaming/AsyncReader/AsyncReader.swift
+++ b/Sources/AsyncStreaming/AsyncReader/AsyncReader.swift
@@ -1,0 +1,69 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if UnstableAsyncStreaming && compiler(>=6.4)
+public import ContainersPreview
+
+/// Reads elements asynchronously from a source.
+///
+/// Adopt ``AsyncReader`` when you need to provide callee-managed buffering,
+/// where the reader controls the buffer and passes a span of elements
+/// to the caller through the `body` closure.
+@available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
+public protocol AsyncReader<ReadElement, ReadFailure>: ~Copyable, ~Escapable {
+  /// The type of elements this reader reads.
+  associatedtype ReadElement: ~Copyable
+
+  /// The error type that reading operations throw.
+  associatedtype ReadFailure: Error
+
+  /// Reads elements from the underlying source and passes them to the provided body closure.
+  ///
+  /// This method asynchronously reads a span of elements from the source,
+  /// then passes them to `body` for processing.
+  ///
+  /// ```swift
+  /// var fileReader: FileAsyncReader = ...
+  ///
+  /// // Read data from a file asynchronously and process it.
+  /// let result = try await fileReader.read { data in
+  ///     guard data.count > 0 else {
+  ///         return
+  ///     }
+  ///     return data
+  /// }
+  /// ```
+  ///
+  /// - Parameter maximumCount: The maximum count of items you're ready
+  ///   to process. Must be greater than zero.
+  /// - Parameter body: A closure that processes a span of read elements
+  ///   and returns a value of type `Return`. When the span is empty,
+  ///   it indicates the end of the stream.
+  /// - Returns: The value the body closure returns after processing the read elements.
+  /// - Throws: An `EitherError` containing either a `ReadFailure` from the read operation
+  ///   or a `Failure` from the body closure.
+  mutating func read<Return: ~Copyable, Failure: Error>(
+    maximumCount: Int,
+    body: (consuming InputSpan<ReadElement>) async throws(Failure) -> Return
+  ) async throws(EitherError<ReadFailure, Failure>) -> Return
+
+}
+
+@available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
+extension AsyncReader where Self: ~Copyable, Self: ~Escapable {
+  /// Reads elements with no upper bound on span size.
+  public mutating func read<Return: ~Copyable, Failure: Error>(
+    body: (consuming InputSpan<ReadElement>) async throws(Failure) -> Return
+  ) async throws(EitherError<ReadFailure, Failure>) -> Return {
+    try await read(maximumCount: .max, body: body)
+  }
+}
+#endif
--- a/Sources/AsyncStreaming/AsyncWriter/AsyncWriter.swift
+++ b/Sources/AsyncStreaming/AsyncWriter/AsyncWriter.swift
@@ -1,0 +1,56 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if UnstableAsyncStreaming && compiler(>=6.4)
+/// Writes elements asynchronously to a destination using a provided buffer.
+///
+/// Adopt ``AsyncWriter`` when you need to provide callee-managed buffering,
+/// where the writer supplies an output span buffer that the caller fills
+/// with elements to write.
+@available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+public protocol AsyncWriter<WriteElement, WriteFailure>: ~Copyable, ~Escapable {
+  /// The type of elements this writer writes.
+  associatedtype WriteElement: ~Copyable
+
+  /// The error type that writing operations throw.
+  associatedtype WriteFailure: Error
+
+  /// Provides a buffer for writing elements to the destination.
+  ///
+  /// The writer supplies an output span that `body` uses to append elements.
+  /// The writer manages the buffer allocation and handles the writing
+  /// operation once `body` completes.
+  ///
+  /// - Parameter body: A closure that receives an `OutputSpan` for appending elements
+  ///   to write. The closure returns a result of type `Return`.
+  ///
+  /// - Returns: The value the body closure returns.
+  ///
+  /// - Throws: An `EitherError` containing either a `WriteFailure` from the write operation
+  ///   or a `Failure` from the body closure.
+  ///
+  /// ## Example
+  ///
+  /// ```swift
+  /// var writer: SomeAsyncWriter = ...
+  ///
+  /// try await writer.write { outputSpan in
+  ///     for item in items {
+  ///         outputSpan.append(item)
+  ///     }
+  ///     return outputSpan.count
+  /// }
+  /// ```
+  mutating func write<Return: ~Copyable, Failure: Error>(
+    _ body: (inout OutputSpan<WriteElement>) async throws(Failure) -> Return
+  ) async throws(EitherError<WriteFailure, Failure>) -> Return
+}
+#endif
--- a/Sources/AsyncStreaming/CallerAsyncWriter/CallerAsyncWriter.swift
+++ b/Sources/AsyncStreaming/CallerAsyncWriter/CallerAsyncWriter.swift
@@ -1,0 +1,51 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if UnstableAsyncStreaming && compiler(>=6.4)
+public import ContainersPreview
+
+/// Writes elements asynchronously from a caller-provided span.
+///
+/// Adopt ``CallerAsyncWriter`` when you need caller-managed buffering,
+/// where the caller provides a span of elements for the writer
+/// to consume.
+@available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
+public protocol CallerAsyncWriter<WriteElement, WriteFailure>: ~Copyable, ~Escapable {
+  /// The type of elements this writer writes.
+  associatedtype WriteElement: ~Copyable
+
+  /// The error type that writing operations throw.
+  associatedtype WriteFailure: Error
+
+  /// Writes a span of elements to the underlying destination.
+  ///
+  /// This method asynchronously writes all elements from the provided span to whatever destination
+  /// the writer represents. The operation may require multiple write calls to complete if the
+  /// writer cannot accept all elements at once.
+  ///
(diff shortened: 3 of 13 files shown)


Sources/_AsyncStreaming/Reader/Array+AsyncReader.swift
```diff
@@ -0,0 +1,85 @@
...
+  ///
+  /// ```swift
+  /// let numbers = [1, 2, 3, 4, 5]
+  /// var reader = numbers.asyncReader()
+  ///
+  /// try await reader.forEach { span in
+  ///     print("Read \(span.count) numbers")
+  /// }
+  /// ```
+  public func asyncReader() -> some AsyncReader<Element, Never> & SendableMetatype {
```
Catfish-Man: Idle thought: if we do the reader/iterator split like we discussed, is this basically the same as makeIterator() if `Array` conformed to `AsyncReader`?

Sources/_AsyncStreaming/Reader/AsyncReader+forEach.swift
```diff
@@ -0,0 +1,104 @@
...
+  /// ```swift
+  /// var fileReader: FileAsyncReader = ...
+  ///
+  /// // Process each chunk of data from the file
+  /// try await fileReader.forEach { chunk in
+  ///     print("Processing \(chunk.count) elements")
+  ///     // Process the chunk
+  /// }
+  /// ```
+  public consuming func forEach<Failure: Error>(
```
Catfish-Man: Concrete benefits of not being API-stable yet :)

Sources/_AsyncStreaming/Reader/AsyncReader+map.swift
```diff
@@ -0,0 +1,100 @@
...
+  #endif
+  mutating func read<Return, Failure>(
+    maximumCount: Int?,
+    body: (consuming Span<MappedElement>) async throws(Failure) -> Return
+  ) async throws(EitherError<Base.ReadFailure, Failure>) -> Return {
+    var buffer = RigidArray<MappedElement>()
+    return try await self.base
+      .read(maximumCount: maximumCount) { (span) throws(Failure) -> Return in
+        guard span.count > 0 else {
+          let emptySpan = InlineArray<0, MappedElement>.zero()
```
Catfish-Man: `EmptyCollection().span` should do the trick

Sources/_AsyncStreaming/Reader/AsyncReader.swift
```diff
@@ -0,0 +1,158 @@
...
+  ///     // Process the data
+  ///     return data
+  /// }
+  /// ```
+  #if compiler(<6.3)
+  @_lifetime(&self)
+  #endif
+  mutating func read<Return, Failure: Error>(
+    maximumCount: Int?,
+    body: (consuming Span<ReadElement>) async throws(Failure) -> Return
```
Catfish-Man: I agree with leaving scatter/gather for a later commit given how tricky it's been

Sources/_AsyncStreaming/Writer/AsyncWriter+AsyncReader.swift
```diff
@@ -0,0 +1,52 @@
...
+  /// try await fileWriter.write(dataReader)
+  /// ```
+  ///
+  /// ## Discussion
+  ///
+  /// This method provides a convenient way to pipe data from one async stream to another,
+  /// automatically handling the iteration and transfer of elements. The operation continues
+  /// until the reader signals completion by producing an empty span.
+  @_lifetime(self: copy self)
+  public mutating func write<ReadFailure: Error>(
```
Catfish-Man: Oh I'd missed seeing this convenience earlier, I like it!

Sources/_AsyncStreaming/Writer/AsyncWriter.swift
```diff
@@ -0,0 +1,162 @@
...
+  /// var writer: SomeAsyncWriter = ...
+  ///
+  /// try await writer.write { outputSpan in
+  ///     for item in items {
+  ///         outputSpan.append(item)
+  ///     }
+  ///     return outputSpan.count
+  /// }
+  /// ```
+  // TODO: EOF should be signaled by providing an empty output span?
```
Catfish-Man: I guess in the tradition of "next() returns nil to indicate end of Sequence" we could also make this optional instead, but using an empty one probably is nicer

Sources/_AsyncStreaming/Writer/AsyncWriter.swift
```diff
@@ -0,0 +1,162 @@
...
+  mutating func write(
+    _ span: Span<WriteElement>
+  ) async throws(EitherError<WriteFailure, AsyncWriterWroteShortError>)
+}
+
+/// An error that indicates the writer was unable to accept all provided elements.
+///
+/// This error is thrown when an async writer signals that it cannot accept any more data
+/// by providing an empty output span, but there are still elements remaining to be written.
+public struct AsyncWriterWroteShortError: Error {
```
Catfish-Man: I forget, was there a reason we couldn't just use "not all the elements have been drained from the span" for this?

(Also do we want InputSpan for this write()?)

Catfish-Man (approved):
Thanks for putting this up :)

Sources/_AsyncStreaming/Writer/AsyncWriter.swift
czechboy0: Though might lead to more accidentally unterminated loops. Optional has the upside that it requires the explicit acknowledgement of nil and probably a guard let that takes care of the terminating condition.

Sources/_AsyncStreaming/Writer/AsyncWriter.swift
```diff
@@ -0,0 +1,162 @@
...
+  ///
+  /// // Write the entire span to a file asynchronously
+  /// try await fileWriter.write(dataBuffer.span)
+  /// ```
+  #if compiler(<6.3)
+  @_lifetime(self: copy self)
+  #endif
+  mutating func write(
+    _ span: Span<WriteElement>
+  ) async throws(EitherError<WriteFailure, AsyncWriterWroteShortError>)
```
czechboy0: Am I reading this right that for the writer, we require both callee and caller owned buffer support, but for the writer, we only require callee?

What's the rationale for having 3/4 of the options, rather than 2/2 or 4/4?

Sources/_AsyncStreaming/Writer/AsyncWriter.swift
```diff
@@ -0,0 +1,162 @@
...
+  @_lifetime(self: copy self)
+  #endif
+  public mutating func write(_ element: consuming WriteElement) async throws(WriteFailure) {
+    // Since the element is ~Copyable but we don't have call-once closures
+    // we need to move it into an Optional and then take it out once. This
+    // also makes the below force unwrap safe
+    var opt = Optional(element)
+    do {
+      try await self.write { outputSpan in
+        outputSpan.append(opt.take()!)
```
czechboy0: This will crash if the output span is full. I guess every place where output span is used should check for remaining capacity before appending even a single element.

It's definitely a more difficult spelling to get right compared to a caller-owned write, where you don't have to explicitly check bounds.

czechboy0 (commented):
Very cool 🙏

Sources/_AsyncStreaming/Writer/AsyncWriter.swift
```diff
@@ -0,0 +1,162 @@
...
+  ///   or a `Failure` from the body closure.
+  ///
+  /// ## Example
+  ///
+  /// ```swift
+  /// var writer: SomeAsyncWriter = ...
+  ///
+  /// try await writer.write { outputSpan in
+  ///     for item in items {
+  ///         outputSpan.append(item)
```
czechboy0: Needs a bounds check.

Sources/AsyncStreaming/AsyncReader/AsyncReader.swift
```diff
@@ -0,0 +1,69 @@
...
+
+#if UnstableAsyncStreaming && compiler(>=6.4)
+public import ContainersPreview
+
+/// Reads elements asynchronously from a source.
+///
+/// Adopt ``AsyncReader`` when you need to provide callee-managed buffering,
+/// where the reader controls the buffer and passes a span of elements
+/// to the caller through the `body` closure.
+@available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
```
Catfish-Man, line 20: 10.14 seems likely incorrect

Sources/AsyncStreaming/AsyncReader/AsyncReader.swift
```diff
@@ -0,0 +1,69 @@
...
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if UnstableAsyncStreaming && compiler(>=6.4)
+public import ContainersPreview
+
+/// Reads elements asynchronously from a source.
+///
+/// Adopt ``AsyncReader`` when you need to provide callee-managed buffering,
```
Catfish-Man, line 17: This shouldn't block landing, but the doc comment here probably shouldn't assume the reader knows about the more general stream protocol concepts

Sources/_AsyncStreaming/AsyncReader/AsyncReader.swift
```diff
@@ -0,0 +1,69 @@
...
+  ///   to process. Must be greater than zero.
+  /// - Parameter body: A closure that processes a span of read elements
+  ///   and returns a value of type `Return`. When the span is empty,
+  ///   it indicates the end of the stream.
+  /// - Returns: The value the body closure returns after processing the read elements.
+  /// - Throws: An `EitherError` containing either a `ReadFailure` from the read operation
+  ///   or a `Failure` from the body closure.
+  mutating func read<Return, Failure: Error>(
+    maximumCount: Int,
+    body: (consuming InputSpan<ReadElement>) async throws(Failure) -> Return
```
Catfish-Man, line 55: Is `consuming` rather than `inout` correct here?

Sources/AsyncStreaming/AsyncWriter/AsyncWriter.swift
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if UnstableAsyncStreaming && compiler(>=6.4)
+/// Writes elements asynchronously to a destination using a provided buffer.
+///
+/// Adopt ``AsyncWriter`` when you need to provide callee-managed buffering,
```
Catfish-Man, line 15: Didn't we want to use the short name for the other one? The "data flows downhill" thing?

Sources/AsyncStreaming/CallerAsyncWriter/CallerAsyncWriter.swift
```diff
@@ -0,0 +1,51 @@
...
+  ///
+  /// ```swift
+  /// var fileWriter: FileAsyncWriter = ...
+  /// let dataBuffer: [UInt8] = [1, 2, 3, 4, 5]
+  ///
+  /// // Write the entire span to a file asynchronously
+  /// try await fileWriter.write(dataBuffer.span)
+  /// ```
+  mutating func write(
+    span: borrowing InputSpan<WriteElement>
```
Catfish-Man, line 48: This has to be `inout` so the stream can give it back nonempty right?

Sources/AsyncStreaming/EitherError.swift
```diff
@@ -0,0 +1,55 @@
...
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if UnstableAsyncStreaming && compiler(>=6.4)
+/// A type-safe wrapper around one of two distinct error types.
+///
+/// Use ``EitherError`` when an operation can fail with errors from two
+/// different sources, such as a read failure and a body closure failure.
+public enum EitherError<First: Error, Second: Error>: Error {
```
Catfish-Man, line 18: Shouldn't block landing, but this'll need to be `@frozen` for embedded (and perf) right?

Sources/_AsyncStreaming/AsyncWriter/AsyncWriter.swift
```diff
@@ -0,0 +1,56 @@
...
+  /// var writer: SomeAsyncWriter = ...
+  ///
+  /// try await writer.write { outputSpan in
+  ///     for item in items {
+  ///         outputSpan.append(item)
+  ///     }
+  ///     return outputSpan.count
+  /// }
+  /// ```
+  mutating func write<Result, Failure: Error>(
```
Catfish-Man: I think this is on me, but: does shadowing the stdlib Result type cause any issues here? I feel like it might

Tests/AsyncStreamingTests/Helpers/Array+Span.swift
```diff
@@ -0,0 +1,43 @@
...
+  init(_ span: Span<Element>) {
+    self.init()
+    for index in span.indices {
+      self.append(span[index])
+    }
+  }
+
+  init(_ inputSpan: borrowing InputSpan<Element>) {
+    self.init()
+    for index in inputSpan.indices {
```
Catfish-Man, line 26: Really highlights how we need a better reserveCapacity that doesn't cause quadratic scaling when accidentally used in a loop

Catfish-Man (changes requested):
General note: it looks like we could use test coverage for two of the protocol types. Want me to whip up a diff for that?

Sources/AsyncStreaming/AsyncReader/AsyncReader.swift
FranzBusch: This is the availability of Span and MutableSpan which seems correct to me. Might need to change in the future.

Sources/AsyncStreaming/AsyncReader/AsyncReader.swift
FranzBusch: I agree. Let's do a doc pass when we are closer to our final shapes.

Sources/AsyncStreaming/AsyncWriter/AsyncWriter.swift
FranzBusch: For now I choose to keep both sides consistent but we need to go through a naming bike shed once we arrive closer to the final shapes. The reason I choose to stay consistent for now is because talking about those types otherwise becomes very confusing.

Sources/_AsyncStreaming/AsyncReader/AsyncReader.swift
FranzBusch: Not sure. Since if there are bytes left they will be dropped. There is no expectation that remaining bytes are stored. At least that is the current expectation.

Sources/AsyncStreaming/CallerAsyncWriter/CallerAsyncWriter.swift
FranzBusch: Similar here the expectation is that all elements are written. Writing short is an error. So arguably this can be `consuming` even

Sources/_AsyncStreaming/AsyncWriter/AsyncWriter.swift
FranzBusch: I changed to `Return` but the shadowing generally doesn't cause problems and we have it in many places in the ecosystem already.

FranzBusch:
> General note: it looks like we could use test coverage for two of the protocol types. Want me to whip up a diff for that?

That was intentional since the two remaining tests require `OutputSpan`s in async contexts which we currently don't have a solution for.

Catfish-Man (approved):

FranzBusch:
I merged over the docs failure which should be fixed by https://github.com/apple/swift-collections/pull/649

PR #406, opened 2026-03-12
Fix availability macro definition for AsyncAlgorithms 1.3

azarovalex:
In Package.swift `AsyncAlgorithms_v1_3` defines the availability domain as `AsyncAlgorithms 1.2` instead of `AsyncAlgorithms 1.3`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,7 +14,7 @@ let AsyncAlgorithms_v1_3 =
 #else
 let AsyncAlgorithms_v1_1 = "AvailabilityMacro=AsyncAlgorithms 1.1:macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0"
 let AsyncAlgorithms_v1_2 = "AvailabilityMacro=AsyncAlgorithms 1.2:macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0"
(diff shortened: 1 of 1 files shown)


josh-arnold-1 (approved):

Package.swift
```diff
@@ -14,7 +14,7 @@ let AsyncAlgorithms_v1_3 =
 #else
```
azarovalex, line 14: @phausler looking at this more it seems that the whole `#else` branch is dead code since `// swift-tools-version: 6.2` is defined for this manifest? It feels like we can remove `#if compiler(>=6.0) && swift(>=6.0)` check unless I'm missing something

phausler:
@swift-ci please test

azarovalex:
@phausler thanks for triggering CI, it seems like it is broken for quite some time tho and failures are not related to thisPR.
I published a new PR that should hopefully fix macOS CI: https://github.com/apple/swift-async-algorithms/pull/409
Linux pipeline seems to have a compiler crash so likely not something we could fix in this repo

phausler:
That linux crash is really gnarly! Have you made a ticket already on the swift repo GH issues? 'cause that is definitely a huge regression

azarovalex:
@phausler 
Okay I found a two line repro for the crash and opened a new issue in the swift repo: https://github.com/swiftlang/swift/issues/88530

I see the crash happens on macOS too in the separate PR where I try to fix macOS CI, I added a one line workaround there: [#409](https://github.com/apple/swift-async-algorithms/pull/409), we'd need to land that PR first to unblock this one.

PR #412, opened 2026-04-28
Make AsyncShareSequence Type Public

AGiersch1804:
# Summary
- Makes the `AsyncShareSequence<Base>` type public.
- Changes the return type of `AsyncSequence.share(bufferingPolicy:)` from the opaque `some AsyncSequence<Element, Failure> & Sendable` type to the concrete `AsyncShareSequence<Self>` type.

This allows dependent codebases to store a sequence that has used the `.share()` modifier.

# Test Plan

- [x] `swift build` succeeds locally


kvaDrug:
Cool, that's what I proposing in the issue: https://github.com/apple/swift-async-algorithms/issues/431

markst:
@phausler care to share your design?

PR #415, opened 2026-05-12
Adds a new `AsyncReader.collect` method

FranzBusch:
Often developers want to collect elements from a reader up to a certain limit. This PR provides a few convenience methods for this.

### DIFF
--- a/Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
+++ b/Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
@@ -1,0 +1,196 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if UnstableAsyncStreaming && compiler(>=6.4)
+
+public import ContainersPreview
+import BasicContainers
+
+@available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+extension AsyncReader where Self: ~Copyable, Self: ~Escapable, ReadElement: ~Copyable {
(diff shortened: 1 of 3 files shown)


Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
```diff
@@ -0,0 +1,194 @@
...
+    /// }
+    /// ```
+    ///
+    /// ## Memory Considerations
+    ///
+    /// Since this method buffers all elements in memory before processing, it should be used
+    /// with caution on large datasets. The `limit` parameter serves as a safety mechanism
+    /// to prevent excessive memory usage.
+    public mutating func collect<Result, Failure: Error>(
+        upTo limit: Int,
```
Catfish-Man: I could see an argument that this might want a default parameter, but I don't think that's critical to decide now

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
```diff
@@ -0,0 +1,194 @@
...
+        body: (consuming InputSpan<ReadElement>) async throws(Failure) -> Result
+    ) async throws(EitherError<ReadFailure, Failure>) -> Result {
+        // TODO: In the future we might want to use a temporary allocation instead
+        // but those don't support async closures yet.
```
Catfish-Man: We should also look into whether we can lower the growth to realloc…

Catfish-Man (approved):
Looks good

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
FranzBusch: We had a lot of arguments in the HTTP APIs around this. Our main argument against a default parameter is that it is impossible to pick the default. The only default to pick is `Int.max` but that makes this vulnerable to remote denial of service attacks if the attack can tricker an essentially unbounded memory collection

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
FranzBusch: I added a `reserveCapacity` to avoid the reallocs.

PR #417, opened 2026-05-13
Rework `AsyncStreaming` protocols to use generic `RangeReplaceableCon…

FranzBusch:
…tainer` buffers

Replace `InputSpan`/`OutputSpan` with generic `Buffer` associated types constrained to `RangeReplaceableContainer` across all four streaming protocols. This sidesteps the limitation that `OutputSpan` (and other `~Escapable` types) cannot be used in async contexts today, while still allowing conforming types to choose a buffer representation optimized for their use case (e.g. `UniqueArray` for heap-backed storage, or a future stack-allocated container for embedded).

Switch callee-owned protocols from `consuming` to `inout` closure semantics, rename `forEachChunk` to `forEachBuffer`, and move `forEach`/`collect` to future directions in the proposal.

Adds test coverage for `AsyncWriter`, `CallerAsyncReader`, and `EitherError`. Updates all documentation to reflect the new API surface.

### DIFF
--- a/Sources/AsyncStreaming/NNNN-async-streaming.md
+++ b/Sources/AsyncStreaming/NNNN-async-streaming.md
@@ -147,7 +147,7 @@ The bite-sized pseudocode:
 
 ```
 protocol AsyncReader {
-    // Callee provides a full buffer; caller processes it
+    // Callee provides a full buffer; caller drains it
     func read(body: (inout buffer) throws -> R) throws -> R
 }
 


Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
```diff
@@ -44,153 +52,41 @@ extension AsyncReader where Self: ~Copyable, Self: ~Escapable, ReadElement: ~Cop
...
-  /// ## Memory Considerations
-  ///
-  /// Since this method buffers all elements in memory before processing, it should be used
-  /// with caution on large datasets. The `limit` parameter serves as a safety mechanism
-  /// to prevent excessive memory usage.
   public mutating func collect<Result, Failure: Error>(
     upTo limit: Int,
     body: (consuming InputSpan<ReadElement>) async throws(Failure) -> Result
-  ) async throws(EitherError<ReadFailure, Failure>) -> Result {
+  ) async throws(EitherError<EitherError<ReadFailure, AsyncReaderLeftOverElementsError>, Failure>) -> Result {
```
Catfish-Man, line 58: `EitherError<EitherError<…>>` wow that's a painful type. Do we need a variadic error instead?

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
```diff
@@ -44,153 +52,41 @@ extension AsyncReader where Self: ~Copyable, Self: ~Escapable, ReadElement: ~Cop
...
+    var collectedBuffer = UniqueArray<ReadElement>()
+    collectedBuffer.reserveCapacity(limit)
     var shouldContinue = true
     do {
       while shouldContinue {
-        try await self.read(
-          maximumCount: limit - buffer.count
-        ) { (span: consuming InputSpan<ReadElement>) in
-          guard span.count > 0 else {
+        try await self.read { (buffer: inout Buffer) throws(AsyncReaderLeftOverElementsError) -> Void in
```
Catfish-Man, line 66: My kingdom for better typed throws inference in closures

Sources/AsyncStreaming/NNNN-async-streaming.md
```diff
@@ -138,32 +138,32 @@ the 2×2 possibility matrix:
...
+API with `inout` access ensures the caller can process elements without taking
+ownership of the buffer, enabling safe reuse.
 
 The bite-sized pseudocode:
 
 ```
 protocol AsyncReader {
-    // Callee provides a full buffer; caller drains it
-    func read(body: (buffer) throws -> R) throws -> R
+    // Callee provides a full buffer; caller processes it
```
Catfish-Man: Wording change since it might be in-place processing?

Sources/AsyncStreaming/NNNN-async-streaming.md
```diff
@@ -175,7 +175,7 @@ question "which one(s) should my type conform to?". We believe that applying
 of situations:
 
 > If you are not sure, pick callee-owned (`AsyncReader`) for read streams and
-> caller-owned (`AsyncWriter`) for write streams.
+> caller-owned (`CallerAsyncWriter`) for write streams.
```
Catfish-Man, line 178: Hm. Weren't we going to give the "good" names to the recommended types?

Sources/AsyncStreaming/NNNN-async-streaming.md
```diff
@@ -494,20 +482,55 @@ patterns, for example:
...
+
+Two common patterns emerge immediately when working with `AsyncReader`: iterating
+over all chunks until the stream ends, and collecting elements into a buffer up
+to a specified limit. We envision convenience extensions for both:
+
+```swift
+extension AsyncReader where Self: ~Copyable, Self: ~Escapable {
+    /// Iterates over all chunks, executing `body` for each buffer until the
+    /// stream ends.
+    public consuming func forEachBuffer<Failure: Error>(
```
Catfish-Man, line 495: Do we want to have the proposal reflect the current state or speculate about the possible syntax sugar?

Tests/AsyncStreamingTests/AsyncReader/AsyncReader+collectTests.swift
```diff
@@ -72,29 +72,17 @@ struct AsyncReaderCollectTests {
 
   @Test
   @available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
-  func collectIntoOutputSpan() async {
-    // TODO: Cannot test this yet since we can't get `InputSpan`s available in async contexts
```
Catfish-Man: Nice to see this go :)

Catfish-Man (approved):
A few minor comments, but basically looks great. Thanks for doing this!

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
FranzBusch: I don't think you can define a variadic enum that generates cases per pack element. I agree this is awful...

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
FranzBusch: IIRC the compiler can do this but we disabled it since it is source breaking. We really should have an upcoming feature that infers this by default again

Sources/AsyncStreaming/NNNN-async-streaming.md
FranzBusch: Yes, this is still like the current state. I wanna revisit the names altogether. I have some better ideas potentially but for now keeping the two sides aligned is easier to think about.

Sources/AsyncStreaming/NNNN-async-streaming.md
FranzBusch: My intention was the current state for now and keep it updated with changes.

Sources/AsyncStreaming/NNNN-async-streaming.md
FranzBusch: actually reverted this back

PR #419, opened 2026-05-14
Fix a few new region isolation errors in `nightly-main`

FranzBusch:
Region isolation got smarter and started diagnosing more potential data races. This PR addresses the new errors. Importantly I had to introduce one `@unchecked Sendable` conformance since we can't yet use `UniqueDeque`


phausler (approved):
Do we have a bug tracking the compiler regression breaking existing code?
The changes look fine here.

FranzBusch:
I filed https://github.com/swiftlang/swift/issues/89141 but it these new errors are real data races and for some of our `@usableFromInline` types we were just missing an explicit `Sendable` conformance.

PR #420, opened 2026-05-15
Add new `pipe` convenience methods for the async readers

FranzBusch:
## Motivation

Often developers want to pipe the contents of an async reader into a writer. This is a common scenario in asynchronous streaming APIs. Readers are modeling sources of data where writers are modeling sinks. As an example imagine reading from a network socket and writing the contents into a file. Importantly, the entire operation needs to uphold backpressure.

## Modifications

This PR adds 2 new `pipe` methods to each reader one for each writer type. Depending on the buffer ownership those methods will the pipe operation is copy free, requires a copy or even requires an intermediate allocation for copying. The method parameter labels are trying to express what the operation does.

## Result

It is very easy for developers now to move elements from readers into writers.

### DIFF
--- a/Sources/AsyncStreaming/AsyncReader/AsyncReader+pipe.swift
+++ b/Sources/AsyncStreaming/AsyncReader/AsyncReader+pipe.swift
@@ -1,0 +1,106 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if UnstableAsyncStreaming && compiler(>=6.4)
+
+import ContainersPreview
+
+@available(macOS 26.2, iOS 26.2, watchOS 26.2, tvOS 26.2, visionOS 26.2, *)
+extension AsyncReader where Self: ~Copyable, Self: ~Escapable, Self.ReadElement: ~Copyable {
+  /// Pipes all elements from this reader into the given writer.
+  ///
+  /// This method consumes the reader and writes all of its elements into the writer's
+  /// destination. It iterates over each buffer the reader produces and hands it to the
+  /// writer until this reader's stream ends.
+  ///
+  /// ## Example
+  ///
+  /// ```swift
+  /// let dataReader: DataAsyncReader = ...
+  /// var fileWriter: FileCallerAsyncWriter = ...
+  ///
+  /// // Copy all data from reader to writer
+  /// try await dataReader.pipe(into: &fileWriter)
+  /// ```
+  ///
+  /// - Parameter writer: A ``CallerAsyncWriter`` to receive the elements. The writer is
+  ///   mutated in place and remains usable after this operation.
+  ///
+  /// - Throws: An error originating from the read or write operations.
+  public consuming func pipe<Writer>(
+    into writer: inout Writer
+  ) async throws(EitherError<ReadFailure, Writer.WriteFailure>)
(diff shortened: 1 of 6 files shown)


Sources/AsyncStreaming/AsyncReader/AsyncReader+pipe.swift
```diff
@@ -0,0 +1,106 @@
...
+  ///
+  /// // Copy all data from reader to writer
+  /// try await dataReader.pipe(into: &fileWriter)
+  /// ```
+  ///
+  /// - Parameter writer: A ``CallerAsyncWriter`` to receive the elements. The writer is
+  ///   mutated in place and remains usable after this operation.
+  ///
+  /// - Throws: An error originating from the read or write operations.
+  public consuming func pipe<Writer>(
```
Catfish-Man, line 38: The functionality here is great, but I *really* don't like this name. We can fix it later (i.e. it shouldn't block landing), but we should find something better. I'd love something like init(consumingContentsOf:), but I can see how that would get awkward if init should take other parameters.

Sources/AsyncStreaming/AsyncReader/AsyncReader+pipe.swift
```diff
@@ -0,0 +1,106 @@
...
+    copyingInto writer: inout Writer
+  ) async throws(EitherError<ReadFailure, Writer.WriteFailure>)
+  where Writer: AsyncWriter & ~Copyable & ~Escapable, Writer.WriteElement == ReadElement {
+    try await self.forEachBuffer { (readerBuffer: inout Buffer) throws(Writer.WriteFailure) in
+      var consumer = readerBuffer.consumeAll()
+      while let firstElement = consumer.next() {
+        var pending: ReadElement? = firstElement
+        do throws(EitherError<Writer.WriteFailure, Never>) {
```
Catfish-Man, line 78: …I can't believe I learned some Swift syntax from this PR 😂

Sources/AsyncStreaming/CallerAsyncReader/CallerAsyncReader+pipe.swift
```diff
@@ -0,0 +1,107 @@
...
+  ///     between the reader and writer. Larger values reduce the number of read and
+  ///     write calls at the cost of memory.
+  ///
+  /// - Throws: An error originating from the read or write operations.
+  public consuming func pipe<Writer>(
+    bufferingInto writer: inout Writer,
+    intermediateCapacity: Int
+  ) async throws(EitherError<ReadFailure, Writer.WriteFailure>)
+  where Writer: CallerAsyncWriter & ~Copyable & ~Escapable, Writer.WriteElement == ReadElement {
+    var buffer = UniqueArray<ReadElement>(minimumCapacity: intermediateCapacity)
```
Catfish-Man, line 86: Another place we need a modernized version of withUnsafeTemporaryAllocation

Catfish-Man (approved):
LGTM although like I said, I'd like to revisit it later

Sources/AsyncStreaming/AsyncReader/AsyncReader+pipe.swift
FranzBusch: Other names I thought about where `fuse(into:)` or even `read(into:)`. I am very open to changing the name but I intentionally did not put it on the writer types as neither an `init` nor a method because I felt it belongs on the producing type and allows nicer chaining. Again everything up for further discussion.

Sources/AsyncStreaming/CallerAsyncReader/CallerAsyncReader+pipe.swift
FranzBusch: Yeah the new `withUnsafeTemporaryAllocation` with `OutputSpan` gets us close but the `OutputSpan` is not available in an `async` context so we can't use it here...

PR #421, opened 2026-05-19
Change all AsyncStreaming availability from 26.2 to SwiftStdlib 5.0

guoye-zhang:
I tested this with the latest compiler snapshot (5-17) and it builds fine. The availability isn't actually required in the test target, presumably because the oldest deployment target it supports is macOS 12.0 aligned.

### DIFF
--- a/Tests/AsyncAlgorithmsTests/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannelTests.swift
+++ b/Tests/AsyncAlgorithmsTests/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannelTests.swift
@@ -962,7 +962,7 @@ final class MultiProducerSingleConsumerAsyncChannelTests: XCTestCase {
     _ = await iterator.next(isolation: nil)
   }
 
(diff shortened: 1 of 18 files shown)


Tests/AsyncAlgorithmsTests/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannelTests.swift
```diff
@@ -971,7 +971,7 @@ final class MultiProducerSingleConsumerAsyncChannelTests: XCTestCase {
         backpressureStrategy: .watermark(low: 2, high: 4)
       )
       let channel = channelAndSource.takeChannel()
-      var source: MultiProducerSingleConsumerAsyncChannel.Source! = consume channelAndSource.source
+      var source: MultiProducerSingleConsumerAsyncChannel.Source? = consume channelAndSource.source
```
guoye-zhang, line 974: I hit a compiler crash which was addressed with this change

FranzBusch (approved):

PR #422, opened 2026-05-20
[AsyncStreaming] Add support for final elements

FranzBusch:
## Motivation

Real transports deliver structured data alongside end-of-stream and need to fuse the last write with the close. HTTP trailers and gRPC status are the obvious cases. Neither can be expressed by an empty-buffer terminator, and both lose H2/H3/QUIC's DATA+END_STREAM coalescing without a fused call.

## Modification

Adds `FinalElement: ~Copyable = Void` as a primary associated type on all four protocols. `AsyncReader` delivers it via a `consuming FinalElement?` closure parameter; `CallerAsyncReader` returns it. Both writers gain a consuming `finish` carrying the last chunk and the payload in one call.

`forEachBuffer`, `collect`, and the `pipe` variants thread the payload through. `collect` becomes consuming and gains a `Void`-final overload returning just the result.

Adds an "Alternatives considered" entry covering what would break without this: HTTP body, gRPC, and fused close on H2/H3/QUIC.

## Result

The four protocols can back HTTP body, gRPC streaming, and similar shapes without giving up the fused close. Default `Void` keeps simple conformers unchanged; `Never` marks infinite streams. Custom and `~Copyable` payloads work end to end.

### DIFF
--- a/Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
+++ b/Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
@@ -27,9 +27,11 @@ extension AsyncReader where Self: ~Copyable, Self: ~Escapable, ReadElement: ~Cop
   /// Collects elements from the reader up to a specified limit and processes them.
   ///
   /// This method continuously reads elements from the async reader, accumulating them in an
-  /// internal buffer until either it reaches the end of the stream or the specified limit.
+  /// internal buffer until either the reader signals end-of-stream (by delivering a
+  /// non-`nil` ``AsyncReader/FinalElement``) or the specified limit is reached.
   /// Once collection completes, it passes the accumulated elements to the provided body
-  /// closure as an `InputSpan` for processing.
+  /// closure as an `InputSpan` for processing, and returns the body's result together
+  /// with the ``AsyncReader/FinalElement``.
   ///
   /// - Parameters:
   ///   - limit: The maximum number of elements to collect. This prevents unbounded memory
@@ -37,7 +39,8 @@ extension AsyncReader where Self: ~Copyable, Self: ~Escapable, ReadElement: ~Cop
   ///   - body: A closure that receives an `InputSpan` containing all collected elements and returns
   ///     a result of type `Result`.
   ///
-  /// - Returns: The value returned by the body closure after processing the collected elements.
+  /// - Returns: A tuple of the body closure's result and the ``AsyncReader/FinalElement``
+  ///   delivered with the terminal chunk.
   ///
   /// - Throws: An `EitherError` wrapping either a read failure (which itself may be an
   ///   ``AsyncReaderLeftOverElementsError`` if the reader produces more elements than the limit),
@@ -48,32 +51,38 @@ extension AsyncReader where Self: ~Copyable, Self: ~Escapable, ReadElement: ~Cop
   /// ```swift
   /// var reader: SomeAsyncReader = ...
   ///
-  /// let processedData = try await reader.collect(upTo: 1000) { span in
+  /// let (processedData, _) = try await reader.collect(upTo: 1000) { span in
   ///     // Process all collected elements
   /// }
   /// ```
-  public mutating func collect<Result, Failure: Error>(
+  // TODO: We should make this method take an inout `RangeReplacableCollection` instead
+  public consuming func collect<Result, Failure: Error>(
     upTo limit: Int,
     body: (consuming InputSpan<ReadElement>) async throws(Failure) -> Result
-  ) async throws(EitherError<EitherError<ReadFailure, AsyncReaderLeftOverElementsError>, Failure>) -> Result {
+  ) async throws(EitherError<EitherError<ReadFailure, AsyncReaderLeftOverElementsError>, Failure>) -> (
+    Result, FinalElement
+  ) {
+    var reader = self
     // TODO: In the future we might want to use a temporary allocation instead
     // but those don't support async closures yet.
     var collectedBuffer = UniqueArray<ReadElement>()
     collectedBuffer.reserveCapacity(limit)
-    var shouldContinue = true
+    var finalElement: FinalElement? = nil
     do {
-      while shouldContinue {
-        try await self.read { (buffer: inout Buffer) throws(AsyncReaderLeftOverElementsError) -> Void in
-          guard buffer.count > 0 else {
-            shouldContinue = false
-            return
+      while finalElement == nil {
+        try await reader.read {
+          (buffer: inout Buffer, final: FinalElement?) throws(AsyncReaderLeftOverElementsError) -> Void in
+          if buffer.count > 0 {
+            if limit - collectedBuffer.count < buffer.count {
+              throw AsyncReaderLeftOverElementsError()
+            }
+            var consumer = buffer.consumeAll()
+            while let element = consumer.next() {
+              collectedBuffer.append(element)
+            }
           }
-          if limit - collectedBuffer.count < buffer.count {
-            throw AsyncReaderLeftOverElementsError()
-          }
-          var consumer = buffer.consumeAll()
-          while let element = consumer.next() {
-            collectedBuffer.append(element)
+          if let final {
+            finalElement = final
           }
         }
       }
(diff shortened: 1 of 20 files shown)


Sources/AsyncStreaming/AsyncWriter/AsyncWriter.swift
```diff
@@ -57,5 +78,51 @@ public protocol AsyncWriter<WriteElement, WriteFailure>: ~Copyable, ~Escapable {
...
+  /// separate parameter) lets the closure compute the payload based on the
+  /// bytes it just wrote — for example a checksum trailer over the body
+  /// content — without needing a scratch buffer.
+  ///
+  /// - Parameter body: A closure that fills the buffer with the final
+  ///   elements and returns the optional ``FinalElement`` payload.
+  /// - Throws: An ``EitherError`` carrying either the underlying write
+  ///   failure or the failure thrown by `body`.
+  consuming func finish<Failure: Error>(
+    body: (inout Buffer) async throws(Failure) -> FinalElement?
```
guoye-zhang: What if the buffer is not big enough for the final chunk?

FranzBusch: I ran into this myself and the only way around it is to not call `finish` if you need to write anything which means there is no way to use this method really. Not sure how to fix this yet.

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
```diff
@@ -48,32 +51,37 @@ extension AsyncReader where Self: ~Copyable, Self: ~Escapable, ReadElement: ~Cop
...
   ///
-  /// let processedData = try await reader.collect(upTo: 1000) { span in
+  /// let (processedData, _) = try await reader.collect(upTo: 1000) { span in
   ///     // Process all collected elements
   /// }
   /// ```
-  public mutating func collect<Result, Failure: Error>(
+  public consuming func collect<Result, Failure: Error>(
     upTo limit: Int,
     body: (consuming InputSpan<ReadElement>) async throws(Failure) -> Result
```
guoye-zhang, line 61: Should the final element be delivered to the closure?

Sources/AsyncStreaming/AsyncReader/AsyncReader.swift
```diff
@@ -28,31 +45,47 @@ public protocol AsyncReader<ReadElement, ReadFailure>: ~Copyable, ~Escapable {
...
-  ///   is empty, it indicates the end of the stream.
+  ///   of read elements together with the optional end-of-stream payload and
+  ///   returns a value of type `Return`. Ownership of `finalElement` transfers
+  ///   to the closure.
   /// - Returns: The value the body closure returns after processing the read elements.
   /// - Throws: An `EitherError` containing either a `ReadFailure` from the read operation
   ///   or a `Failure` from the body closure.
   mutating func read<Return: ~Copyable, Failure: Error>(
-    body: (inout Buffer) async throws(Failure) -> Return
+    body: (inout Buffer, consuming FinalElement?) async throws(Failure) -> Return
```
guoye-zhang, line 87: Nested Optionals are hard to deal with. What about an enum `{ case incomplete, case final(FinalElement) }`?

FranzBusch: For now I have decided against that. The nested optional case is really only when a reader/writer defines its final element as optional.

Tests/AsyncStreamingTests/AsyncReader/AsyncReader+pipe.swift
```diff
@@ -18,136 +18,69 @@ import Testing
...
   func pipeIntoCopiesAllElements() async throws {
     let reader = UniqueArrayAsyncReader(
       storage: UniqueArray(capacity: 5, copying: [1, 2, 3, 4, 5])
     )
-    var writer = UniqueArrayCallerAsyncWriter()
+    let writer = UniqueArrayCallerAsyncWriter()
 
-    try await reader.pipe(into: &writer)
-
-    #expect(writer.storage.count == 5)
```
FranzBusch, line 30: There are a few tests where I had to drop the expectations since `pipe` is now `consuming`. However, I have added them back in a follow up PR https://github.com/apple/swift-async-algorithms/pull/425. I needed a few more fundamental building blocks to properly write those tests.

guoye-zhang (approved):
We could simplify `CallerAsyncWriter` into a single method that takes a buffer and an optional final element, since the code to write the buffer can usually be shared between those methods.

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
```diff
@@ -48,32 +51,38 @@ extension AsyncReader where Self: ~Copyable, Self: ~Escapable, ReadElement: ~Cop
   /// ```swift
   /// var reader: SomeAsyncReader = ...
   ///
-  /// let processedData = try await reader.collect(upTo: 1000) { span in
+  /// let (processedData, _) = try await reader.collect(upTo: 1000) { span in
   ///     // Process all collected elements
   /// }
   /// ```
-  public mutating func collect<Result, Failure: Error>(
+  // TODO: We should make this method take an inout `RangeReplacableCollection` instead
```
gjcairo, line 58: What's keeping us from doing this now? Also shouldn't it be a `RangeReplaceableContainer`?

Sources/AsyncStreaming/AsyncReader/AsyncReader+pipe.swift
```diff
@@ -13,92 +13,150 @@
...
+          pending = nil
+          while writerBuffer.freeCapacity > 0 {
+            guard let element = consumer.next() else { return }
+            writerBuffer.append(element)
           }
         }
+      } catch {
+        switch error {
+        case .first(let writeFailure): throw writeFailure
+        case .second: fatalError("Unreachable")
```
gjcairo, line 158: If the second error in `EitherError` is `Never`, can't we call `unwrap()` and have the type system realise it only throws the first? Or is it not that smart that we need this?

Sources/AsyncStreaming/AsyncReader/AsyncReader.swift
```diff
@@ -17,42 +17,74 @@ public import ContainersPreview
 /// Adopt ``AsyncReader`` when you need callee-managed buffering,
 /// where the reader controls the buffer and passes it to the caller
 /// through the `body` closure.
+///
+/// ## Signaling end of stream
+///
+/// The reader signals end-of-stream by passing a non-`nil` value for the
+/// `finalElement` parameter of the `body` closure. This call may also carry a
```
gjcairo, line 24: This would allow empty buffers being received but not meaning end of stream, right? Is that a concern?

Tests/AsyncStreamingTests/AsyncReader/AsyncReader+forEachTests.swift
```diff
@@ -65,7 +66,7 @@ struct AsyncReaderforEachBufferTests {
     let reader = UniqueArrayAsyncReader(storage: UniqueArray(capacity: 3, copying: [1, 2, 3]))
 
     do {
-      try await reader.forEachBuffer { (_) throws(TestError) -> Void in
+      _ = try await reader.forEachBuffer { (_) throws(TestError) -> Void in
```
gjcairo, line 69: Nit: do we want to annotate `forEachBuffer` as `@discardableResult`?

gjcairo (commented):
LGTM overall just a few questions.

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
FranzBusch: I intend to change this API in a future PR to be spelled `collect(into: &buffer)` and that should return the `FinalElement` as well.

FranzBusch:
> We could simplify `CallerAsyncWriter` into a single method that takes a buffer and an optional final element, since the code to write the buffer can usually be shared between those methods.

I thought about this but I think this is worse for two reasons:
1. `finish` is `consuming` right now and makes it impossible to produce more writes after finishing
2. The ergonomics become a lot worse when always having an optional final element on `write`

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
FranzBusch: I want to keep the scope of the PR limited to one change at a time

Sources/AsyncStreaming/AsyncReader/AsyncReader+pipe.swift
FranzBusch: Sadly no. The problem is that would require extensions like this

```swift
extension EitherError where First == Never {
  public func unwrap() throws(Second) {
    switch self {
    case .first:
      fatalError()
    case .second(let second):
      throw second
    }
  }
}

extension EitherError where Second == Never {
  public func unwrap() throws(First) {
    switch self {
    case .first(let first):
      throw first
    case .second:
      fatalError()
    }
  }
}
```

which result in ambiguity at the call site since you can't overload on the thrown type

Sources/AsyncStreaming/AsyncReader/AsyncReader.swift
FranzBusch: Some protocols have it as valid to send empty buffers for liveliness probes so this would be a supported pattern.

Tests/AsyncStreamingTests/AsyncReader/AsyncReader+forEachTests.swift
FranzBusch: I wanted to see more adoption before doing that.

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
gjcairo: Sure. Should it not be a `Container` though instead of a `Collection`? Out of curiosity, you don't have to update the comment.

gjcairo (approved):

Sources/AsyncStreaming/AsyncReader/AsyncReader+collect.swift
FranzBusch: Ah yeah it should. My bad :D

guoye-zhang:
> 1. `finish` is `consuming` right now and makes it impossible to produce more writes after finishing

Yeah not being consuming is unfortunate, and would necessitate a separate function.

> 2. The ergonomics become a lot worse when always having an optional final element on `write`

That's somewhat debatable. Suppose we are writing a middleware that transforms the body:

```
func write(buffer: Buffer) {
    let processed = // process body
    self.next.write(buffer: processed)
}

func finish(buffer: Buffer, finalElement: Void) {
    let processed = // process body
    self.next.finish(buffer: processed, finalElement: ())
}
```

Implementations have to either duplicate code or make both functions call a common helper

guoye-zhang:
Are we planning to swap AsyncWriter and CallerAsyncWriter in this PR or a later PR?

Also should we just remove `collect` conveniences for now since they are not the final shape that we want?

FranzBusch:
> Also should we just remove `collect` conveniences for now since they are not the final shape that we want?

I will change them in a follow up.

> Are we planning to swap AsyncWriter and CallerAsyncWriter in this PR or a later PR?

We won't swap the naming here but we need to swap in the HTTP API proposal repo.

Catfish-Man (approved):
I remain extremely conflicted on this at a conceptual level. The PR itself seems fine, and I think it's fine to go ahead with it so we can experiment with it, but it still feels like we're prioritizing networking-shaped use-cases over general ones, not without some cost.

On the other hand, representing infinite streams in the type system with FinalElement == Never is so elegant that it almost wins me over just on the strength of that.

Anyway, we can try this out and see how it goes.

PR #424, opened 2026-05-26
[AsyncStreaming] Add bidirectional adapters for the async writers

FranzBusch:
## Motivation

It is quite common to have one type of writer and wanting to adapt to the other type of writer.

## Modifications

This PR adds two adapters into either direction between the writers to make conversion seamless.

## Result

It is easy to go from one type of writer to the other type.

### DIFF
--- a/Sources/AsyncStreaming/AsyncWriter/AsyncWriterCallerAsyncWriterAdapter.swift
+++ b/Sources/AsyncStreaming/AsyncWriter/AsyncWriterCallerAsyncWriterAdapter.swift
@@ -1,0 +1,98 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Async Algorithms open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if UnstableAsyncStreaming && compiler(>=6.4)
+public import ContainersPreview
+
+/// A ``CallerAsyncWriter`` that is implemented in terms of an
+/// ``AsyncWriter``.
+///
+/// Each ``write(buffer:)`` call drains the caller's buffer through one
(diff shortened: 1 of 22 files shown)


Sources/AsyncStreaming/AsyncWriter/AsyncWriterCallerAsyncWriterAdapter.swift
```diff
@@ -0,0 +1,98 @@
...
+/// or more ``AsyncWriter/write(_:)`` closures on the underlying writer.
+/// When the underlying writer's buffer fills before the caller's empties,
+/// the adapter loops with another closure call to continue draining.
+///
+/// The adapter introduces no buffer of its own — elements move directly
+/// from the caller-supplied buffer into the underlying writer's
+/// closure-supplied buffer. The underlying writer's deferred-flush
+/// behavior, if any, is preserved.
+@available(macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
+public struct AsyncWriterCallerAsyncWriterAdapter<
```
Catfish-Man, line 28: [standard comment about names goes here]

we'll figure it out later

Sources/AsyncStreaming/AsyncWriter/AsyncWriterCallerAsyncWriterAdapter.swift
```diff
@@ -0,0 +1,98 @@
...
+    while let head = consumer.next() {
+      var pending: WriteElement? = head
+      do throws(EitherError<WriteFailure, Never>) {
+        try await self.underlying.write {
+          (innerBuffer: inout Underlying.Buffer) async throws(Never) -> Void in
+          if case .some(let element) = pending.take() {
+            innerBuffer.append(element)
+          }
+          while innerBuffer.freeCapacity > 0 {
+            guard let element = consumer.next() else { return }
```
Catfish-Man, line 57: Getting all of this to lower to memcpy calls when the types are appropriate is going to be a fascinating exercise at some point. I never quite got it to for `AsyncBytes`

Catfish-Man (approved):
Looks good :) nice functionality

PR #425, opened 2026-05-26
[AsyncStreaming] Introduce a `MultiProducerSingleConsumerChannel` and a duplex

FranzBusch:
## Motivation

When working with async reader and writer one often needs "root" readers and writers. A multi producer single consumer channel can provide the foundation for such a reader and writer pair. Furthermore, developers often want a connected duplex pair of readers and writers for testing or in-memory purposes.

## Modifications

This PR brings over the existing `MultiProducerSingleConsumerAsyncChannel` with adaptions to handle reader and writer like interfaces and leverage unstable types such as `UniqueDeque`. Furthermore, this PR also brings a `DuplexStream` that sets up two connected pairs of MPSC channels.

I intentionally copied and modified the MPSC implementation since I don't want to carry `#if` conditionals in an existing used implementation. Once we stabilize this module we can evaluate how to best proceed.

## Result

Makes it easier for developers to work with readers and writers. In particular, makes testing a lot easier.

### DIFF
--- a/Sources/AsyncStreaming/DuplexChannel/DuplexAsyncChannel.swift
+++ b/Sources/AsyncStreaming/DuplexChannel/DuplexAsyncChannel.swift
@@ -28,7 +28,7 @@ public import ContainersPreview
 ///   readerA <──────────────────────────────────────── writerB
 /// ```
 ///
-/// The four handles are independent ``~Copyable`` values so each can be
+/// The four handles are independent `~Copyable` values so each can be
 /// sent to its own task without an intermediate decomposition step.
 ///
 /// Each direction applies backpressure independently using the configured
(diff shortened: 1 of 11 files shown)


Sources/AsyncStreaming/DuplexChannel/DuplexAsyncChannel.swift
```diff
@@ -0,0 +1,536 @@
...
+  /// - Returns: The value returned from `body`.
+  @inlinable
+  public static func withDuplex<Result: ~Copyable, BodyFailure: Error>(
+    of elementType: Element.Type = Element.self,
+    withFinalElement finalElementType: FinalElement.Type,
+    throwing failureType: Failure.Type = Never.self,
+    backpressureStrategy: BackpressureStrategy,
+    isolation: isolated (any Actor)? = #isolation,
+    body: (
+      consuming sending Writer,
```
guoye-zhang, line 96: Why 2 pairs? One pair is more commonly used and you can always build 2 from 1?

guoye-zhang (approved):

Sources/AsyncStreaming/DuplexChannel/DuplexAsyncChannel.swift
FranzBusch: The MPSCChannel is a single pair the duplex is just two pairs connected bi-directionally. Both of them are important and yes one can be build using the other. In fact this PR does that. Having a connected duplex stream is very useful for many testing scenarios.

Catfish-Man:
Sent you a patch on Slack to fix the tests building. With that I'm ok with landing this

guoye-zhang:
Merged in #428

PR #426, opened 2026-05-29
Make MultiProducerSingleConsumerAsyncChannel internals compatible with library evolution

### LINKED ISSUE #405: MultiProducerSingleConsumerAsyncChannel breaks library evolution (BUILD_LIBRARY_FOR_DISTRIBUTION=YES) in 1.1.x
## Summary

`swift-async-algorithms` 1.1.x is incompatible with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` (Swift library evolution mode). Every version in the 1.1.x line (1.1.0 through 1.1.3) fails to compile when used as a dependency of an XCFramework built with library evolution enabled. The 1.0.x line (including 1.0.4) compiles cleanly under the same conditions.

This is a blocking issue for any team distributing a binary SDK as an XCFramework that depends on `swift-async-algorithms`.

---

## Environment

| Property | Value |
|----------|-------|
| Xcode | 26.3 |
| Swift | 6.2.4 |
| Platform | iOS |
| Minimum deployment target | iOS 16.0 |
| swift-async-algorithms | 1.1.3 (reproduced on 1.1.0, 1.1.1, 1.1.2) |
| Build flag | `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` |

---

## How to Reproduce

Add `swift-async-algorithms` 1.1.x as a dependency to any Swift package and archive with library evolution enabled:

```bash
xcodebuild archive \
  -workspace MySDK.xcworkspace \
  -scheme MySDK \
  -destination "generic/platform=iOS" \
  -archivePath build/MySDK.xcarchive \
  BUILD_LIBRARY_FOR_DISTRIBUTION=YES \
  SKIP_INSTALL=NO
```

---

## Errors

All 17 errors originate from a single file introduced in 1.1.0:

```
Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannel+Internal.swift
```

### Type 1 — `@inlinable` on a class designated initializer (1 error)

```
error: initializer for class '_Storage' is '@inlinable' and must delegate to another initializer
```

Under library evolution, a class's stored properties are accessed through resilient accessors. A designated initializer must initialize those properties directly and therefore cannot be `@inlinable`.

**Location:** `_Storage.init(backpressureStrategy:)` — line 122.

### Type 2 — `switch consume self._state` on a non-`@frozen` struct (16 errors)

```
error: 'consume' can only be used to partially consume storage
```

`_StateMachine` is a non-`@frozen` struct. Under library evolution its stored properties are accessed through resilient accessors rather than direct storage. The `consume` operator requires direct ownership of storage and cannot be applied to `self._state` through a resilient accessor.

All 16 affected locations are `@inlinable mutating func` methods on `_StateMachine`:

| Line | Method |
|------|--------|
| 552 | `setOnTerminationCallback(sourceID:callback:)` |
| 596 | `cancelOnTerminationCallback(sourceID:)` |
| 628 | `finish(sourceID:error:)` |
| 694 | `cancelSource(sourceID:)` |
| 723 | `registerConsumer(continuation:)` |
| 782 | `cancelConsumer(continuation:)` |
| 835 | `next(for:)` |
| 888 | `send(contentsOf:)` — first overload |
| 967 | `send(contentsOf:)` — second overload |
| 1057 | `send(_:)` |
| 1107 | `addSource()` |
| 1155 | `removeSource(sourceID:)` |
| 1209 | `cancelSource(sourceID:)` — second overload |
| 1288 | `nextFromConsumer()` |
| 1373 | `nextFromProducer(sourceID:)` |
| 1458 | `cancelNextFromProducer(sourceID:)` |

> **Note:** The `switch consume sendResult` patterns in `MultiProducerSingleConsumerAsyncChannel.swift` are not affected — those operate on a local variable, not a stored property, and are valid under library evolution.

---

## Question

Was the incompatibility with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` / library evolution considered when `MultiProducerSingleConsumerAsyncChannel` was introduced in 1.1.0? Is there a plan to address this in an upcoming release?

The `1.0.x` line remains the only usable version for teams building XCFrameworks today.

adityasingh2400:
`MultiProducerSingleConsumerAsyncChannel`, introduced in 1.1.0, fails to compile when the consuming module is built with library evolution enabled (`BUILD_LIBRARY_FOR_DISTRIBUTION=YES`, i.e. `-enable-library-evolution`). This blocks every team that distributes a binary SDK as an XCFramework depending on `swift-async-algorithms`; the 1.0.x line is currently the only usable version for them.

Two classes of error come from `Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannel+Internal.swift`:

- `'consume' can only be used to partially consume storage` — the `@inlinable mutating func` methods on the non-`@frozen` `_StateMachine` struct (and the `_InternalBackpressureStrategy` enum) do `switch consume self._state` / `switch consume self`. Under library evolution a non-frozen type's stored properties are reached through resilient accessors, which do not vend the direct ownership `consume` requires.
- `initializer for class '_Storage' is '@inlinable' and must delegate to another initializer`, plus the related `'let' property ... may not be initialized directly` errors on the `@inlinable` initializers of the `~Copyable` payload structs. Under library evolution an `@inlinable` initializer of a resilient type may only assign through a delegating `self.init`/`self = ...`, not set stored properties directly.

These reproduce on 1.1.0–1.1.3; `main` is still affected.

Changes:

- Mark the internal (`@usableFromInline`) types whose stored storage is `consume`d or directly initialized in `@inlinable` members as `@frozen`, so their layout is fixed and direct storage access / `consume` is permitted under library evolution: `_StateMachine`, `_State` and its `Channeling`/`SourceFinished`/`Finished` payloads, `_InternalBackpressureStrategy` and its `_Watermark`, the `ResumeElement`/`ResumeElementAndProducers`/`ResumeFailure` wrappers, and `SendableConsumeOnceBox`. These are all already `@usableFromInline` internal types, so freezing them is not a public-API or source-stability change.
- Mark `_TinyArray` and its `Storage` enum `@frozen` for the same reason (they back the frozen payloads).
- Change `_Storage.init(backpressureStrategy:)` from `@inlinable` to `@usableFromInline`, since an `@inlinable` designated class initializer must delegate under library evolution and this one sets a stored property directly.

No public API changes and no behavioral changes — only resilience-attribute additions on internal types and one inlinability downgrade on an internal initializer.

Testing:

- `swift build --target AsyncAlgorithms -Xswiftc -enable-library-evolution` now succeeds (it fails on `main` with the errors above).
- `swift build` (normal, non-evolution) succeeds.
- `swift test --filter MultiProducerSingleConsumer` passes — 37 tests, 0 failures.

Fixes #405.

### DIFF
--- a/Sources/AsyncAlgorithms/Internal/_TinyArray.swift
+++ b/Sources/AsyncAlgorithms/Internal/_TinyArray.swift
@@ -26,8 +26,10 @@
 /// It supports arbitrary many elements but if only up to one ``Element`` is stored it does **not** allocate separate storage on the heap
 /// and instead stores the ``Element`` inline.
 @usableFromInline
+@frozen
 struct _TinyArray<Element> {
   @usableFromInline
+  @frozen
   enum Storage {
     case one(Element)
     case arbitrary([Element])
--- a/Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannel+Internal.swift
+++ b/Sources/AsyncAlgorithms/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannel+Internal.swift
@@ -16,8 +16,10 @@ import Synchronization
 @available(AsyncAlgorithms 1.1, *)
 extension MultiProducerSingleConsumerAsyncChannel {
   @usableFromInline
+  @frozen
   enum _InternalBackpressureStrategy: Sendable, CustomStringConvertible {
     @usableFromInline
+    @frozen
     struct _Watermark: Sendable, CustomStringConvertible {
       /// The low watermark where demand should start.
       @usableFromInline
@@ -118,7 +120,7 @@ extension MultiProducerSingleConsumerAsyncChannel {
     @usableFromInline
     let _stateMachine: Mutex<_StateMachine>
 
-    @inlinable
+    @usableFromInline
     init(
       backpressureStrategy: _InternalBackpressureStrategy
     ) {
@@ -505,6 +507,7 @@ extension MultiProducerSingleConsumerAsyncChannel {
 extension MultiProducerSingleConsumerAsyncChannel._Storage {
   /// The state machine of the channel.
   @usableFromInline
+  @frozen
   struct _StateMachine: ~Copyable {
     /// The state machine's current state.
     @usableFromInline
@@ -1358,6 +1361,7 @@ extension MultiProducerSingleConsumerAsyncChannel._Storage {
       ///   destructively pattern-matching an enum case with multiple payloads
       ///   that include a `~Copyable` generic wrapper type (`Disconnected`).
       @usableFromInline
+      @frozen
       struct ResumeElement: ~Copyable, Sendable {
         @usableFromInline let continuation: UnsafeContinuation<Element?, Error>
         @usableFromInline var element: Disconnected<Element>
@@ -1375,6 +1379,7 @@ extension MultiProducerSingleConsumerAsyncChannel._Storage {
       ///
       /// - Note: See `ResumeElement` for the reason behind this wrapper.
       @usableFromInline
+      @frozen
       struct ResumeElementAndProducers: ~Copyable, Sendable {
         @usableFromInline let continuation: UnsafeContinuation<Element?, Error>
         @usableFromInline var element: Disconnected<Element>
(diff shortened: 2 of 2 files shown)


FranzBusch:
This package does not support compiling with library evolution mode enabled. If you use this inside an XCFramework then you should only import it using `internal import`

adityasingh2400:
Thanks for the context, that makes sense. I had not clocked that the package intentionally does not support library-evolution mode. Using internal import inside the XCFramework is the right call here. Appreciate the quick explanation.

PR #427, opened 2026-06-06
#425 plus a build fix

Catfish-Man:
I can't push to Franz's branch and he's OoO, so this is his branch + one fix to make the test suite pass


Catfish-Man:
Ugh. git.

PR #436, opened 2026-06-24
[AsyncStreaming] Redesign AsyncReader.collect to take a target container

FranzBusch:
Replace `collect(upTo:body:)` with three target-based overloads: `collect(into:)`, `collect(exactlyInto:)`, and `collect(into:maximumSize:)`. Add `AsyncReaderInsufficientElementsError` for the exact-match shortfall case.


Catfish-Man:
Yeah this looks great. Thank you!

Catfish-Man (approved):

Issue #391, opened 2025-12-27
Add next(isolation:) support to async sequences


alex-waldron:
[SE-0421: Generalize effect polymorphism for AsyncSequence and AsyncIteratorProtocol](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0421-generalize-async-sequence.md) introduced `next(isolation:)` to `AsyncIteratorProtocol`. Poking around this package, I realized that most iterators are yet to implement the isolated version of next. Is there a specific reason why?

allenhumphreys:
@alex-waldron Definitely others-- `merge`, `combineLatest`, and `zip` all suffer the crash as well. While this can be fixed in this library, the root cause may be in the compiler and has to do with the `Error` type that was added, like you said. So the solution proposed in #441 would be a pretty quick fix for consumers of this library. But long term, it would seem like dealing with Error type on the API would be better. 

The crash only occurs when the library consumer has a minimum deployment target >= iOS 18.

```swift
@Test
func test_merge2_neverPlusError_triggersNeverCast() async {
    struct ContrivedError: Error {}
    do {
        for try await _ in merge(
            AsyncStream<Int> { _ in },
            AsyncThrowingStream<Int, Error> { $0.finish(throwing: ContrivedError()) },
        ) {}
        Issue.record("merge(2) did not throw or trap")
    } catch {
        // No crash: Failure did not resolve to Never for merge(2).
    }
}
@Test
func test_merge3_neverPlusError_triggersNeverCast() async {
    struct ContrivedError: Error {}
    do {
        for try await _ in merge(
            AsyncStream<Int> { _ in },
            AsyncStream<Int> { _ in },
            AsyncThrowingStream<Int, Error> { $0.finish(throwing: ContrivedError()) },
        ) {}
        Issue.record("merge(3) did not throw or trap")
    } catch {
        // No crash: Failure did not resolve to Never for merge(3).
    }
}
@Test
func test_combineLatest2_neverPlusError_triggersNeverCast() async {
    struct ContrivedError: Error {}
    do {
        for try await _ in combineLatest(
            AsyncStream<Int> { _ in },
            AsyncThrowingStream<Int, Error> { $0.finish(throwing: ContrivedError()) },
        ) {}
        Issue.record("combineLatest(2) did not throw or trap")
    } catch {
        // No crash: Failure did not resolve to Never for combineLatest(2).
    }
}
@Test
[...]

PR #440, opened 2026-06-30
[Merge] Add isolated next for AsyncMerge2Sequence

julianiaccopucci:
## Summary
- Add an explicit `next(isolation:)` implementation to `AsyncMerge2Sequence.Iterator` so Swift 6 iteration does not fall back to the standard-library compatibility bridge.
- Add regression coverage for mixed throwing/non-throwing `merge` inputs in both argument orders.

## Motivation
With Swift 6 typed `AsyncSequence.Failure`, `for await` can call `next(isolation:)`. Without an explicit implementation, `AsyncMerge2Sequence.Iterator` can use the default bridge through legacy `next()`, which force-casts thrown errors to the inferred `Failure` type. For mixed throwing and non-throwing inputs, that can trap when a later input throws while the merged sequence is inferred as non-throwing.

This is related to #391.

## Test plan
- `swift test --filter TestMerge2`
- `swift test --filter TestMerge`
- `swift test`

### DIFF
--- a/Sources/AsyncAlgorithms/Merge/AsyncMerge2Sequence.swift
+++ b/Sources/AsyncAlgorithms/Merge/AsyncMerge2Sequence.swift
@@ -102,6 +102,10 @@ extension AsyncMerge2Sequence {
     public mutating func next() async rethrows -> Element? {
       try await internalClass.next()
     }
+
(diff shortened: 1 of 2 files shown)


julianiaccopucci:
Closing this for now while I tighten the regression coverage for the Swift 6  availability path. I’ll reopen or submit an updated PR once the local tests are in the right shape.

PR #449, opened 2026-08-24
[AsyncStreaming] Propagate cancellation to late readers

### LINKED ISSUE #448: DuplexAsyncChannel/MPSC cancellation hang
I'm using the DuplexAsyncChannel and finding that sometimes the code just hangs. The following is what I believe to be a reproduction of that hang. 

```
#if UnstableAsyncStreaming && compiler(>=6.4)

import AsyncStreaming

#if canImport(Glibc)
import Glibc
#elseif canImport(Darwin)
import Darwin
#endif

@available(macOS 15.0, iOS 18.0, tvOS 18.0, visionOS 2.0, *)
@main
struct MinimalRepro {
  static func main() async {
    await withTaskGroup(of: Void.self) { group in
      // A child added to an already-cancelled group is cancelled at creation, so
      // the read inside `forEachBuffer` starts out cancelled — no polling, no
      // sleeping, and no dependence on which task the executor runs first.
      group.cancelAll()
      group.addTask {
        await DuplexAsyncChannel<Int, Void, Never>.withDuplex(
          of: Int.self,
          backpressureStrategy: .watermark(low: 1, high: 2)
        ) { writerA, readerA, writerB, readerB in
          _ = consume readerA
          _ = consume writerB
          // Dropped unfinished, so the forward direction stays open and empty:
          // the read cannot complete and has to park.
          _ = consume writerA

          print("forEachBuffer on an already-cancelled reader must throw CancellationError.")
          print("If this process is still alive, it did not: it is spinning. Ctrl-C to stop.")
          // Flushed because the spin below never returns, and a piped stdout would
          // otherwise never be drained.
          fflush(nil)

          do {
            // Loops until a read reports a non-`nil` final element. The lost
            // cancellation means that never happens, and because each read
            // completes without suspending, this pegs a core and stops observing
            // cancellation. `body` is never invoked: `forEachBuffer` skips it for
            // an empty buffer, and every chunk here is empty.
            _ = try await readerB.forEachBuffer { _ in
              print("body invoked")
            }
            print("forEachBuffer returned")
          } catch {
            print("OK: forEachBuffer threw \(error)")
          }
        }
      }
    }
  }
}

#endif
```

thliu21:
## Summary

- Resume a late reader continuation with `CancellationError` when cancellation finishes the channel before the continuation is installed.
- Preserve the normal source-finished path.
- Add deterministic regression coverage using a task that starts cancelled.

## Root cause

For an already-cancelled task, the cancellation handler can run before `suspendRead()` installs its continuation. `cancelRead()` marks the channel as finished due to cancellation, but the late continuation was previously resumed successfully. The read loop then observed an empty, non-terminal result, causing `forEachBuffer` to continue indefinitely.

The state machine now distinguishes cancellation from normal source completion and resumes the late continuation with the cancellation error only for the cancellation path.

## Validation

Using the Swift 6.4 release toolchain on macOS:

- The focused regression test on the unfixed baseline failed with the expected two issues.
- The focused regression test passed after the fix: 1 test passed.
- The AsyncStreaming test suite passed: 87 tests passed.
- `git diff --check` passed.

The full package suite and Linux, Windows, and WebAssembly matrices were not run locally; those remain for the repository PR CI.

Fixes #448.


FranzBusch (approved):

FranzBusch (changes requested):
Upon second review this PR is not enough. We can't just throw `CancellationError` on all `finished` states. We have to track why we entered finished and only if the reader was cancelled throw the `CancellationError`. If the source finished regularly then we can't throw this error. We should test this. Also we don't have to re-throw the error on subsequent reads. Once we threw the error or returned the final element once we should just precondition since it violates the contract.

thliu21:
Thanks for the second look. I updated the state machine to distinguish a pending reader cancellation from an already-consumed terminal result. Only pending cancellation now resumes with CancellationError; normal completion does not. Repeated reads after cancellation, a final element, or a source failure now precondition, with coverage for each path. The Swift 6.3.2 full suite passes locally (358/358), along with strict format and diff checks. The new Swift 6.4-specific tests are pushed, but the current Actions run is waiting for workflow approval: https://github.com/apple/swift-async-algorithms/actions/runs/33357550882

thliu21:
Follow-up pushed in `f6ef36a`.

This closes the remaining cancellation race across both late-reader orderings: whether cancellation is recorded before the reader continuation is installed, or a producer takes and resumes the continuation immediately before the cancellation handler runs, the pending cancellation is delivered exactly once. Normal finish still delivers `finalElement`, and a subsequent read after cancellation, final-element delivery, or a read-side failure preconditions.

I also updated EOS callers and tests to terminate on `finalElement` rather than an empty buffer, and aligned the Duplex reader documentation.

Local validation with Swift 6.4-dev (2026-08-28 snapshot):
- focused trait-enabled Debug: 37 XCTest + 20 Swift Testing passed
- full trait-enabled Release: 358 XCTest + 92 Swift Testing passed
- full trait-enabled Debug is blocked by an `OwnershipModelEliminator` compiler assertion, reproduced unchanged on unmodified `origin/main`

The current GitHub Actions workflow still requires approval, so no CI result is being claimed.

Tests/AsyncStreamingTests/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannelTests.swift
```diff
@@ -11,12 +11,14 @@
 
 #if UnstableAsyncStreaming && compiler(>=6.4)
 
-import AsyncStreaming
+@testable import AsyncStreaming
```
FranzBusch: Can we please avoid this. We should only test the public interface

Tests/AsyncStreamingTests/MultiProducerSingleConsumerChannel/MultiProducerSingleConsumerAsyncChannelTests.swift
```diff
@@ -75,7 +77,148 @@ struct MultiProducerSingleConsumerAsyncChannelTests {
...
+          }
+        }
+      }
+      await group.waitForAll()
+    }
+  }
+
+  @available(macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
+  @Test
+  func readAvailableDeliversPendingCancellation() async {
```
FranzBusch: I know this test is not implementable without the testable import so can we just drop it

FranzBusch:
@thliu21 thanks. This looks good besides the testable import. If you can remove that I am happy to merge the PR

thliu21:
Thanks! I removed the `@testable` import and dropped `readAvailableDeliversPendingCancellation`; the remaining regression coverage uses only the public API. Swift 6.4-dev Release passes locally—37 XCTest + 19 Swift Testing focused, and 358 XCTest + 91 Swift Testing for the full trait-enabled suite. Strict format and diff checks also pass. The prior CI run’s only failure was in the untouched `AsyncAlgorithms/FlatMapLatest` code, and that test passes locally.

